PR #22, opened 2017-08-16
fixed pagination style bug

### LINKED ISSUE #17: pagination组件样式
pagination组件，在点到最后一页后，下一页 按钮在禁用时鼠标移上去还有蓝色背景，不知是不是原本就是这样设计的

mrxf:
Fixed #17  pagination style bug.

And if the Mixin `button-color` in button componet as a common mixin would be better.

### DIFF
--- a/src/components/pagination/style/index.less
+++ b/src/components/pagination/style/index.less
@@ -189,6 +189,7 @@
   &-disabled {
     &:hover {
       border-color: @border-color-base;
+      background-color: @disabled-bg;
       a {
         color: @disabled-color;
         cursor: not-allowed;


vthinkxie:
Thanks for your PR, but index.less file is sync with [ant design react](https://github.com/ant-design/ant-design/blob/master/components/pagination/style/index.less) which should not be changed here.

It seems that class `ant-pagination-item-active` was added incorrectly when last index or first index is reached

https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/src/components/pagination/nz-pagination.component.ts#L40

Wait for you to PR again, thanks.

PR #31, opened 2017-08-16
docs(message): unify API type bool to Boolean

SangKa:
The most component pages used the `Boolean`  and the Message component use the `bool`.
So make them uniform.


trotyl:
Shouldn't it better to use `boolean` to keep consistency of type annotation?

SangKa:
I agree with @trotyl , but need to change all of them in the showcase.

PR #36, opened 2017-08-16
feat(module:root): make nz-root optional

### LINKED ISSUE #34: Make <nz-root> component optional
## I'm submitting a...

<pre><code>
[ ] Bug report  
[x] Feature request
[ ] Documentation issue or request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Support request => Please do not submit support request here
</code></pre>

## Current behavior

As suggested in doc, `<nz-root>` component must be listed at outermost layer of the app, which seems inconvenient for user.

#### High-level goal:

Do not require `<nz-root>` in component template.

#### Detailed design:

- [x] Add optional configuration param in `forRoot()` method.
- [x] Introduce a separate `InjectionToken` for that configuration param.
- [x] Use a dynamic component to apply global styles.
- [x] Use a `APP_INITIALIZER` to apply font styles.

## Environment

<pre><code>
Angular version: X.Y.Z

ng-zorro-antd version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] IE version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

trotyl:
Closes #34 .

### API in detail:

For `NgZorroAntdModule.forRoot()` method, there is an optional parameter in signature `{ extraFontName: string, extraFontUrl: string }`, like:

```typescript
const options = { extraFontName: 'some-name', extraFontUrl: 'some-url' };

imports: [
  NgZorroAntdModule.forRoot(options)
]
```

It would be same as using `<nz-root nzExtraFontName="some-name" nzExtraFontUrl="some-url">`.

Meanwhile, user can manually provide an `NZ_ROOT_CONFIG` if they don't want to import the entire lib.

### DIFF
--- a/src/components/root/nz-root.module.ts
+++ b/src/components/root/nz-root.module.ts
@@ -1,12 +1,28 @@
-import { NgModule } from '@angular/core';
+import { NgModule, OnDestroy, ComponentRef, ComponentFactoryResolver, Inject, Optional, Injector, APP_INITIALIZER } from '@angular/core';
+import { CommonModule, DOCUMENT } from '@angular/common';
 import { NzRootComponent } from './nz-root.component';
-import { CommonModule } from '@angular/common';
+import { NzRootStyleComponent } from './nz-root-style.component';
+import { NZ_ROOT_CONFIG, createNzRootInitializer } from './nz-root-config';
 
 @NgModule({
-  exports     : [ NzRootComponent ],
-  declarations: [ NzRootComponent ],
-  imports     : [ CommonModule ]
+  exports         : [ NzRootComponent ],
+  declarations    : [ NzRootComponent, NzRootStyleComponent ],
+  imports         : [ CommonModule ],
+  entryComponents : [ NzRootStyleComponent ],
+  providers       : [
+    { provide: APP_INITIALIZER, multi: true, useFactory: createNzRootInitializer, deps: [DOCUMENT, [new Optional(), NZ_ROOT_CONFIG]] },
+  ],
 })
+export class NzRootModule implements OnDestroy {
+  private styleHostComponent: ComponentRef<NzRootStyleComponent>;
 
-export class NzRootModule {
+  constructor(@Inject(DOCUMENT) _document: Document, injector: Injector, resolver: ComponentFactoryResolver) {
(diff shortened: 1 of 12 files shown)


trotyl:
The coverage has increased, but since the original coverage is fake (most of the source files didn't being count into), it has wrong result in the report.

vthinkxie:
original coverage count is not correct.

src/components/root/nz-root.module.ts
```diff
@@ -1,12 +1,28 @@
...
+  providers       : [
+    { provide: APP_INITIALIZER, multi: true, useFactory: createNzRootInitializer, deps: [DOCUMENT, [new Optional(), NZ_ROOT_CONFIG]] },
+  ],
 })
+export class NzRootModule implements OnDestroy {
+  private styleHostComponent: ComponentRef<NzRootStyleComponent>;
 
-export class NzRootModule {
+  constructor(@Inject(DOCUMENT) _document: Document, injector: Injector, resolver: ComponentFactoryResolver) {
+    const componentFactory = resolver.resolveComponentFactory(NzRootStyleComponent);
```
vthinkxie: user may insert `nz-root` manually , it may need to add judgement here.

vthinkxie:
docs needed after nz-root become optional.

src/components/root/nz-root.module.ts
trotyl: It's already being resolved on `NzRootComponent`: 

> this.nzExtraFontName && this.nzExtraFontUrl && **!this.options**.

trotyl:
Should it be:

+ Make `<nz-root>` the default option, with this being an alternative. Or
+ Make this the default option, with `<nz-root>` being an alternative. Or
+ Make this the suggested way, with `<nz-root>` being a deprecated option. Or
+ Make this the only documented option, with `<nz-root>` not mentioned anywhere?

And should we change the way that showcase app uses?

vthinkxie:
what about removing `nz-root` from the document and tell users that it has been deprecated on the changelog page, and the showcase app should also change.

trotyl:
Sounds good, would update it soon.

PR #240, opened 2017-09-06
build(packaging): produce angular package format compatible code

trotyl:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[x] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Application (the showcase website) / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: closes #144, fixes #85 

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

Issue #85, opened 2017-08-18
Not support for Rollup in AOT mode


miao17game:
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Support request => Please do not submit support request here
</code></pre>

## Current behavior

When i try to rollup for the AOT mode, i find that it's impossible to build with ng-zorro.

the error message is below here:

<pre><code>
> ngc -p tsconfig-aot.json && rollup -c rollup-config.js

⚠️   'default' is imported from external module 'rollup' but never used

'default' is not exported by 'node_modules\@angular\common\@angular\common.es5.js'
'default' is not exported by 'node_modules\@angular\core\@angular\core.es5.js'
�   Error: Cannot call a namespace ('moment')
node_modules\ng-zorro-antd\src\release\datepicker\nz-datepicker.component.js (17:26)
15:     this._value = null;
16:     this._today = new Date();
17:     this._selectedMonth = moment(this.nzValue).month();
                              ^
18:     this._selectedYear = moment(this.nzValue).year();
19:     this._selectedDate = moment(this.n
</code></pre>

## What is the motivation / use case for changing the behavior?

the problem is the 'moment', it's imported in time-picker component in this format:
<pre><code>
import * as moment from 'moment'
</code></pre>
but it's not support in AOT mode.
if i change the format into 
<pre><code>
import moment from 'moment'
</code></pre>
,no errors will be called up, but still some other errors in moment.
if i remove the NG-ZORRO from my angular4 project, there is nothing wrong.
And there is no problems in JIT mode.

I don't think change the source by myself is a good way to resolve this problem.
I need help.

## Environment

<pre><code>
Angular version: 4.2.4

ng-zorro-antd version: 0.5.0

Browser:
- [x] Chrome (desktop) version 59
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] IE version XX
 
For Tooling issues:
- Node version: XX  
- Platform: Windows7 

Others:

</code></pre>

miao17game:
well, i have gotten a new way to resolve this problem, the is the plugins section in my rollup-config:
<pre><code>
plugins: [
        nodeResolve({ jsnext: true, module: true }),
        commonjs({
            namedExports: {
[...]

PR #315, opened 2017-09-17
no need to fix moment. introduce moment in right way

LinBoLen:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #85 #169 #240

```
[ ] Yes
[x] No
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -43,20 +43,6 @@ if (format === 'es') {
   })
 }
 
-function fixMoment() {
-  return {
-    name: 'fix-moment-import',
-    transform(content) {
-      if (!content.includes(`import * as moment from 'moment'`)) { return }
-
-      return content.replace(
-          `import * as moment from 'moment'`,
-          `import moment from 'moment'`
-      )
-    },
-  }
-}
-
 let input
 let file
 
@@ -81,7 +67,7 @@ export default {
   },
   exports: 'named',
   name: 'ngZorro.antd',
-  plugins: [fixMoment(), resolve()],
+  plugins: [resolve()],
   external: Object.keys(globals),
   globals,
 }
--- a/src/components/calendar/nz-calendar.component.ts
+++ b/src/components/calendar/nz-calendar.component.ts
@@ -10,7 +10,7 @@ import {
   EventEmitter, HostBinding
 } from '@angular/core';
 
-import * as moment from 'moment';
+import moment from 'moment';
 import { Moment } from 'moment';
 import 'moment/locale/zh-cn';
 
--- a/src/components/datepicker/nz-datepicker.component.ts
+++ b/src/components/datepicker/nz-datepicker.component.ts
@@ -8,7 +8,7 @@ import {
   ViewChild,
   HostBinding, OnInit
 } from '@angular/core';
-import * as moment from 'moment';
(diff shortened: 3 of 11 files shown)


LinBoLen:
it seems that it will cause typescript lint error in angular cli project.
but we can superress the error

vthinkxie:
@trotyl any idea?

trotyl:
This would require user to have `allowSyntheticDefaultImports` on their own `tsconfig`, would be a breaking change (even is small).

trotyl:
@LinBoLen Can you provide what your error is, working fine in my local env and CI.

trotyl:
Another issue is that `default` auto-fallback it not available in closure compiler, even we didn't officially support it now, it should be considered for all Angular library (in not introducing compatibility issue deliberately).

LinBoLen:
- never mind the lint error, it comes from webstorm default typescript service, this is just one of thousand lint error.
- `allowSyntheticDefaultImports ` is used for developer work on ng-zorro-antd.  no user
- closure compiler auto-fallback without `allowSyntheticDefaultImports` in typescript

trotyl:
The `.d.ts` is provided along with the library distribution, it would show TypeScript error for no default export, unless user also have set `allowSyntheticDefaultImports` in their `tsconfig`.

`allowSyntheticDefaultImports` would not change generated result, and closure didn't care about `.d.ts` at all, whatever set in `tsconfig` won't help.

trotyl:
You can have a look of [official usage](https://github.com/angular/material2/blob/master/src/material-moment-adapter/adapter/moment-date-adapter.ts#L17-L19), namespace export are always required (as long as we don't want to break user), and our in-build process is much simpler than this. (And it's OK since we just have one library)

trotyl:
I guess you may have some misunderstandings about `allowSyntheticDefaultImports`.

`allowSyntheticDefaultImports` is a type-checking specific option (like `allowUnusedLabels` or `noImplicitAny`), only defines the type-checking rule, results in no change of generated files (none of `.js`, `.d.ts` or `.map` would be affected).

This kind of options have no propagation behavior, even if it compiles in your workspace, does not mean it can work in it's consumers'. (eg. `https://github.com/angular/angular/issues/17863`, `https://github.com/angular/angular/issues/17522`, or `https://github.com/angular/angular/issues/17800`, not want to pollute them so didn't use links) 

What you need may be covered in `https://github.com/Microsoft/TypeScript/issues/16093`.

LinBoLen:
but it seems that just use `import moment from moment`. it works

trotyl:
That's because currently we don't have any type dependency of the namespace import.

You can try with:

```typescript
import * as moment from 'moment';
export const t = moment();
```

Then the generated `.d.ts` file will be:

```typescript
import * as moment from 'moment';
export declare const t: moment.Moment;
```

So that our way of importing would be propagated to our consumers.

In `moment.d.ts`, there're hundred of types, once any of them be used implicitly in our code, the users will be break.

trotyl:
I suggest that just making the same Pull Request to `material2` and they can answer it better (if they want).

PR #393, opened 2017-09-29
docs(API): add note on nzDefaultValue

tigercosmos:
Add note about `nzDefaultValue` in  `nz-slider` `API`

r? @vthinkxie

### DIFF
--- a/src/showcase/nz-demo-slider/nz-demo-slider.html
+++ b/src/showcase/nz-demo-slider/nz-demo-slider.html
@@ -110,7 +110,7 @@
       </tr>
       <tr>
         <td>nzDefaultValue</td>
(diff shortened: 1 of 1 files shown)


vthinkxie:
@wilsoncook

tigercosmos:
@wilsoncook What I want to express is, I expect `nzDefaultValue` should be the default value no matter `ngModel` has value, however the result is not.

For example:
```
 <Slider defaultValue={[20, 50]} [(ngModal)]="sliderValue">
```
which `sliderValue= [0, 100]`, and the result of slider would be ranged from 0 to 100, not as the expected 20 to 50. It would only be 20 to 50 if `sliderValue` does not set value.

That is this PR want to talk about.

wilsoncook:
@tigercosmos AFAIK, to keep the value under a range, should use `min` and `max` option, the `defaultValue` only means something initialize value: if the `ngModal` haven't been initialized, use `defaultValue` as an alternative.

Issue #409, opened 2017-10-06
NzModalService's method 'onOk' cannot find 'this'


tigercosmos:
## I'm submitting a...

<pre><code>
[X] Bug report  
[ ] Feature request
[ ] Documentation issue or request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Support request => Please do not submit support request here
</code></pre>

## Current behavior

`app.component.ts`
```
import { Component } from '@angular/core';
import { NzModalService } from 'ng-zorro-antd';

@Component({
  selector: 'my-app',
  template: `
    <button nz-button [nzType]="'info'" (click)="showConfirm()">
      <span>确认对话框</span>
    </button>
  `,
  styles  : []
})
export class AppComponent {
  showConfirm = () => {
    console.log('outside: ', this);
    this.confirmServ.confirm({
      title: '您是否确认要删除这项内容',
      content: '<b>一些解释</b>',
      onOk() {
        console.log('inside: ', this);
      },
      onCancel() {
      }
    });
  }

  constructor(private confirmServ: NzModalService) {
  }
}
```

and the console get
```
outside:  Object { confirmServ: Object, showConfirm: AppComponent/this.showConfirm() }  app.component.ts:29:25   app.component.ts:34:33

inside:  undefined
```

as the result, `onOk` method cannot find `this`, which is `AppComponent`  itself. This issue will cause the methods in `AppComponent` inside `onOk` do not work.

## Environment

<pre><code>
Angular version: 4.3.6

ng-zorro-antd version: 0.5.1

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] IE version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

tigercosmos:
`let that = this;` can solve the issue, but I am not sure if it is a bug or caused by the native JS mechanism.

```
showConfirm = () => {
    let that = this;
    console.log('outside: ', that);
    this.confirmServ.confirm({
      title: '您是否确认要删除这项内容',
      content: '<b>一些解释</b>',
      onOk() {
        console.log('inside: ', that);
      },
      onCancel() {
      }
    });
  }
```

HsuanXyz:
Uses arrow functions instead of shorthand syntax.

```typescript
showConfirm = () => {
    console.log('outside: ', this);
    this.confirmServ.confirm({
[...]

PR #410, opened 2017-10-08
docs(showcase): use better syntax in NzModal

tigercosmos:
According to #409, update `nzModal`'s showcase in document with better syntax.
r? @vthinkxie


wilsoncook:
@tigercosmos Actually in my opinion, no one is better than another between:
```
{
  onOk() {},
}
```
and
```
{
onOk: () => {}
}
```
depend on personal habits.

tigercosmos:
@wilsoncook Have you seen #409?

tigercosmos:
the original one would cause bugs at some time.
that's why change to the new one.

PR #538, opened 2017-11-06
README: Add note for document

tigercosmos:
Add note on `README`. 
User can find older version document as well as the latest one.


tigercosmos:
@vthinkxie r?

vthinkxie:
I suggest not to show the old version, since they are not under maintain any longer, it may confuse the users and bring issues that has been fixed in latest version.

PR #2570, opened 2018-12-02
More used nz icon

JLHwung:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A
Bundling ng-zorro-antd to a webpack dll should imports all used icons

## What is the new behavior?
Some icons are missing

```
[ ] Yes
[x] No
```

## Other information
The `NZ_ICONS_USED_BY_ZORRO` is a non-compulsory contract between the component authors and the bundler. It is not easy and straightforward for authors to write `nz-icon type="foo"` and then just remember to check the `nz-icon` service whether the icon `FooOutline` is registered. What's more, the typical usage of `ng-zorro-antd` will not bail in most situations where a full ant-desigin/icons-angular is served, thus it is more difficult for author to recognize such kinds of error without testing against the other situations.

That is why I find out there are 5 components missing explicit icon import --- people just don't know about there is, mentally, such kind of `after-use-nz-icon-type` hook.

### Proposed solution
A perfect solution will not land without an architect change of ant-design/icons-angular. But we can kindly remind authors to register icon usage in our PR template. A checklist, i.e.
```
### component template
- [ ] Added icon import on `components/icon/nz-icon.service.ts`
```
will help on this scenario.


wendzhue:
@JLHwung Thanks for you hard work! May I ask do you include icons that are only used in demos?

JLHwung:
> May I ask do you include icons that are only used in demos?

No. I searched the term`nz-icon` with file mask `*.component.html` in the scope of `path/to/ng-zorro-antd/components`, so I don't think the icons used in demos would be matches.

PR #2703, opened 2018-12-28
Fix issue#2586

Wetinee:
Add 'upload' part to make NzI18nService support fr_FR.

### DIFF
--- a/components/i18n/languages/fr_FR.ts
+++ b/components/i18n/languages/fr_FR.ts
@@ -14,8 +14,8 @@ export default {
     filterConfirm: 'OK',
     filterReset: 'Réinitialiser',
     emptyText: 'Aucune donnée',
-      selectAll: 'tout sélectionner',
-      selectInvert: 'sélectionnez inverser',
+    selectAll: 'tout sélectionner',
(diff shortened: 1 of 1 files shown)


components/i18n/languages/fr_FR.ts
```diff
@@ -14,6 +14,8 @@ export default {
     filterConfirm: 'OK',
     filterReset: 'Réinitialiser',
     emptyText: 'Aucune donnée',
+      selectAll: 'tout sélectionner',
```
wendzhue: Indent

Wetinee: OK and thanks for your advice.

wendzhue: Oh. And the first character should be uppercase. Didn't noticed that. Sorry. 🤣

Wetinee: My mistake. See if it's ok now.

wendzhue:
Thanks for submitting this pr. Related problem was closed by https://github.com/NG-ZORRO/ng-zorro-antd/pull/2737.

PR #3006, opened 2019-03-01
docs(module:calendar): add sample usage of `TemplateRef`

dodobelieve:
docs(Calendar): add sample usage of TemplateRef

Add sample usage of TemplateRef,  show date in the tempate.

### DIFF
--- a/components/calendar/doc/index.zh-CN.md
+++ b/components/calendar/doc/index.zh-CN.md
@@ -34,7 +34,7 @@ registerLocaleData(zh);
   <div *dateCell>Foo</div>
 </nz-calendar>
 <!-- 传入 TemplateRef 的方式 -->
-<ng-template #dateCellTpl>Bar</ng-template>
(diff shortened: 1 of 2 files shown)


components/calendar/doc/index.zh-CN.md
```diff
@@ -34,7 +34,7 @@ registerLocaleData(zh);
   <div *dateCell>Foo</div>
 </nz-calendar>
 <!-- 传入 TemplateRef 的方式 -->
-<ng-template #dateCellTpl>Bar</ng-template>
+<ng-template #dateCellTpl let-date><span>{{ date.getDate() }}</span></ng-template>
```
HsuanXyz: `index.en-US.md` also needs to be this change

trotyl: How about using `{{ date | date:'d'}}`? `DatePipe` is `pure: true` and avoids the re-invocation. Also calling function in template is not recommended.

dodobelieve: sure,it's better.

Issue #1651, opened 2018-06-13
 Accessibility


Huangdongrong:
## What problem does this feature solve?
组件是否考虑对ARIA的支持, 如果框架走国际路线，必须要支持 Accessibility，国外有些地区法律法规需要有Accessibility支持。

## What does the proposed API look like?
Angular Material Design 的组件基本会支持Accessibility
https://material.angular.io/components/paginator/overview

vthinkxie:
we will support universal render first, and will take aria into consideration then.

jamesliauw:
@vthinkxie  现在的v1.2是否已完全支持SSR了呢？

mikerobi:
I've started working on aria support for tabs, will probably work on menus next.

wendzhue:
@mikerobi Feel free to make pull requests! 😆

mikerobi:
@wendzhue I'm working on it.

Is it appropriate to add an "aria" commit type to commit-message.json?

How do you feel about warning messages if aria properties are missing?  I've currently got it set up to warn in dev mode.

mikerobi:
I'm sorry i didn't know commits to my own branch would add messages here.  I will leave of the the issue id for my in progress commits.

PR #3145, opened 2019-03-24
fix(module:time-picker): make keyboard navigation possible

jimmytheneutrino:
Tabbing away from time-picker overlay was not possible: It wrapped
around and usually ended up in browser's address bar. Also, once you
selected time-picker wrapper span, there was no way to open the overlay
via keyboard.
This fixes this. Tabbing away from overlay now focuses on time-picker
wrapper span. Pressing enter on the wrapper span opens the overlay.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Tabbing away from time-picker overlay was not possible: It wrapped
around and usually ended up in browser's address bar. Also, once you
selected time-picker wrapper span, there was no way to open the overlay
via keyboard.

Issue Number: N/A

## What is the new behavior?
Tabbing away from overlay now focuses on time-picker
wrapper span. Pressing enter on the wrapper span opens the overlay.

```
[ ] Yes
[x] No
```

## Other information
The behavior (both current and new) can be observed on the main demo page (npm run site:start).

### DIFF
--- a/components/time-picker/nz-time-picker.component.html
+++ b/components/time-picker/nz-time-picker.component.html
@@ -6,7 +6,8 @@
   [placeholder]="nzPlaceHolder || ('TimePicker.placeholder' | nzI18n)"
   [(ngModel)]="value"
   readonly="readonly"
-  (click)="open()">
+  (click)="open()"
+  (keyup.enter)="open()">
 <span class="ant-time-picker-icon">
   <i nz-icon nzType="clock-circle"></i>
 </span>
@@ -50,7 +51,9 @@
     [opened]="nzOpen"
     [nzClearText]="nzClearText"
     [nzAllowEmpty]="nzAllowEmpty"
+    (keyup.enter)="close()"
     [(ngModel)]="value">
   </nz-time-picker-panel>
+  <span tabindex="0" (focus)="close()" class="nz-tab-catching-span"></span>
 </ng-template>
 
--- a/components/time-picker/nz-time-picker.component.spec.ts
+++ b/components/time-picker/nz-time-picker.component.spec.ts
@@ -10,6 +10,7 @@ import { NzTimePickerModule } from './nz-time-picker.module';
 
 import { registerLocaleData } from '@angular/common';
 import zh from '@angular/common/locales/zh';
+import { dispatchFakeEvent } from 'ng-zorro-antd/core';
 registerLocaleData(zh);
 
 describe('time-picker', () => {
@@ -87,6 +88,27 @@ describe('time-picker', () => {
       expect(testComponent.openChange).toHaveBeenCalledTimes(3);
       expect(testComponent.open).toBe(true);
     });
+    it('should open and close on enter', fakeAsync(() => {
+      testComponent.date = new Date('2018-11-11 11:11:11');
+      testComponent.open = false;
+      fixture.detectChanges();
+      testComponent.nzTimePickerComponent.inputRef.nativeElement.dispatchEvent(
+        new KeyboardEvent('keyup', { key: 'enter' })
+      );
+      fixture.detectChanges();
+      tick(500);
+      fixture.detectChanges();
+      expect(testComponent.open).toBe(true);
+      const panel = overlayContainer.getContainerElement().querySelector('nz-time-picker-panel');
+      expect(panel).toBeTruthy();
+      if (panel) {
+        panel.dispatchEvent(new KeyboardEvent('keyup', { key: 'enter' }));
+      }
+      fixture.detectChanges();
+      tick(500);
+      fixture.detectChanges();
+      expect(testComponent.open).toBe(false);
+    }));
     it('should clear work', fakeAsync(() => {
       fixture.detectChanges();
       testComponent.date = new Date('2018-11-11 11:11:11');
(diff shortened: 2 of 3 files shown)


components/time-picker/nz-time-picker.component.html
```diff
@@ -49,7 +50,9 @@
     [opened]="nzOpen"
     [nzClearText]="nzClearText"
     [nzAllowEmpty]="nzAllowEmpty"
+    (keyup.enter)="close()"
     [(ngModel)]="value">
   </nz-time-picker-panel>
+  <span tabindex=0 (focus)="close()" class="nz-tab-catching-span"></span>
```
wilsoncook: Why class "nz-tab-catching-span"?

wilsoncook:
Hi @jimmytheneutrino , we're very grateful for your work! It would be better to provide the actual usage scenario for this PR (as i think there is no need to handle the enter key on `<input>` only, unless we provide the full set of keyboard processing mechanisms such as "left/right arrow to select a specified time value").

components/time-picker/nz-time-picker.component.html
wilsoncook: Also the syntax `tabindex=0` should be `tabindex="0"`

jimmytheneutrino: > Why class "nz-tab-catching-span"?
So that it is easier to handle in the future (if there is need) plus it is a bit self-documenting.

> Also the syntax tabindex=0 should be tabindex="0"
Ok. I will fix that.

jimmytheneutrino:
@wilsoncook 
This onEnter is for the case when the user inputs values by hand using digits on the keyboard.
IMHO the user does not need any arrow key bindings when they already can input just digits.

jimmytheneutrino:
@wilsoncook 
Hi! Do you need something else fixed?

jimmytheneutrino:
@wilsoncook Could you please tell me what else needs to be done?

wenqi73:
@jimmytheneutrino Thanks for your PR, this belongs to `a11y` (#1651) which contains full set of keyboard processing as @wilsoncook said, not just `OnEnter`. It will be helpful that if you provide a full solution of this component.

jimmytheneutrino:
@wenqi73 @wilsoncook 
I understand that the full a11y solution would be preferable. Unfortunately, this seems to be a large task and I do not have time to do all of its nuances.

However, isn't a partial improvement better than no improvement? Please see the videos:
https://imgur.com/a/IhI9R6i .

I think that just this little bit could be a huge improvement in usability for some users.

components/time-picker/nz-time-picker.component.html
```diff
@@ -50,7 +51,9 @@
     [opened]="nzOpen"
     [nzClearText]="nzClearText"
     [nzAllowEmpty]="nzAllowEmpty"
+    (keyup.enter)="close()"
     [(ngModel)]="value">
   </nz-time-picker-panel>
+  <span tabindex="0" (focus)="close()" class="nz-tab-catching-span"></span>
```
wenqi73:
```suggestion
 
```
In theory the focus point should be moved to time picker panel when pressing `Tab`, it should not close the panel.

components/time-picker/nz-time-picker.component.html
```diff
@@ -6,7 +6,8 @@
   [placeholder]="nzPlaceHolder || ('TimePicker.placeholder' | nzI18n)"
   [(ngModel)]="value"
   readonly="readonly"
-  (click)="open()">
+  (click)="open()"
+  (keyup.enter)="open()">
```
wenqi73: You should add the tests about your changes.

components/time-picker/nz-time-picker.component.html
jimmytheneutrino: You mean the same way as it is done in the react version: https://ant.design/components/time-picker/ ?

But I am sure this is a UX bug in that version. This behaviour would be completely unusable with a keyboard. You need to press tab about a hundred times to get anything done there. In reality, choosing time on the time-picker is not needed if the user is already using the keyboard.

components/time-picker/nz-time-picker.component.html
jimmytheneutrino: Added tests.

components/time-picker/nz-time-picker.component.html
wenqi73: The right way is: Users use `Tab` to toggle HOUR/MINITE/SECOND/INPUT selection, use `UP_ARROW` or `DOWN_ARROW` to select number.

jimmytheneutrino: Really? In order to input the time this way in the worst case the user needs 12 + 30 + 30 keystrokes.
Why would anyone do it if they can just input time with numbers, press tab and go to the next input on the form? The way you propose the user would just have to press tab 3 times more to skip through those selections.

The way I see its usage is that: 
a) there are some users who use mouse to enter time - and the time panel is great for that,
b) but there are some 'more experienced' users who want to do it faster -- and they could use keyboard for that -- 6 strokes+tab -- might take less than 2 seconds.
Imagine a receptionist somewhere who does it every day, hundreds of times.

Of course, one could implement a special form for the case b), but if the standard component could do it, why not try with a standard one?

wenqi73: We need to support to cover everywhere of the panel only use keyboard, I mean use `Tab` to toggle when the panel is open, if user want to do it faster they can use 6 strokes + `Enter` + `Tab` to switch to next element, the `Tab` should not just close panel.

vthinkxie:
@wenqi73 plz check this again

vthinkxie:
As there is no update for a long time, the PR will be closed.
If you have any questions, feel free to open another PR, thanks a lot.

PR #3209, opened 2019-04-02
fix(module:collapse): fix glitch when opening collapse in Angular 7

Timebutt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently the `collapse` component glitches when opening, it does not glitch when closing. This can be closely observed if a larger text is put into the collapse (Lorum Ipsum for instance). This was not an issue in previous versions of `ng-zorro` (when comparing versions on the documentation website).

Issue Number: N/A

## What is the new behavior?
The new behaviour removes the `display: none` set by the trigger. In Angular 7 having this property causes the glitch issue.

```
[ ] Yes
[x] No
```

## Other information

Q: the `collapseMotion` still has a `hidden` state, that I didn't remove right now, but I am wondering why it is there in the first place? Shouldn't there just be a `expanded` and `collapsed` state and that's it?

### DIFF
--- a/components/core/animation/collapse.ts
+++ b/components/core/animation/collapse.ts
@@ -4,7 +4,7 @@ import { AnimationCurves } from './animation';
 export const collapseMotion: AnimationTriggerMetadata = trigger('collapseMotion', [
   state('expanded', style({ height: '*' })),
   state('collapsed', style({ height: 0, overflow: 'hidden' })),
-  state('hidden', style({ height: 0, display: 'none' })),
+  state('hidden', style({ height: 0 })),
   transition('expanded => collapsed', animate(`150ms ${AnimationCurves.EASE_IN_OUT}`)),
   transition('expanded => hidden', animate(`150ms ${AnimationCurves.EASE_IN_OUT}`)),
   transition('collapsed => expanded', animate(`150ms ${AnimationCurves.EASE_IN_OUT}`)),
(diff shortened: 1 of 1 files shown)


components/core/animation/collapse.ts
```diff
@@ -4,7 +4,7 @@ import { AnimationCurves } from './animation';
 export const collapseMotion: AnimationTriggerMetadata = trigger('collapseMotion', [
   state('expanded', style({ height: '*' })),
   state('collapsed', style({ height: 0, overflow: 'hidden' })),
-  state('hidden', style({ height: 0, display: 'none' })),
+  state('hidden', style({ height: 0 })),
```
wendzhue: This would result in two borders between two panels.

X4gwA5grIr

components/core/animation/collapse.ts
```diff
@@ -4,7 +4,7 @@ import { AnimationCurves } from './animation';
 export const collapseMotion: AnimationTriggerMetadata = trigger('collapseMotion', [
   state('expanded', style({ height: '*' })),
```
wendzhue:
```suggestion
  state('expanded', style({ height: '*', display: '*' })),
  state('collapsed', style({ height: 0, overflow: 'hidden' })),
  state('hidden', style({ height: 0, display: 'none' })),
```

wendzhue:
To be honest I think it should be a bug of Angulars' - different behaviors in different browsers.

kfrancois:
The border could be fixed by adding `border: none;` to the hidden state, but is this a viable solution?

Issue #3539, opened 2019-06-05
nz-tabset click the lastest tab always has 64px margin-right


NearZXH:
### Reproduction link
[https://stackblitz.com/edit/ng-zorro-antd-tab-bug](https://stackblitz.com/edit/ng-zorro-antd-tab-bug)

### Steps to reproduce
1. set `@tabs-scrolling-size: 0px` in `theme.less`
2. add enough tabs so the `prev` and `next` can occur
3. click the last tab you can see until the lastest tab is occured, you will find a 64 margin-right at the end

### What is expected?
no margin-right at the end.

### What is actually happening?
always has 64px margin-right at the end

---
1. First of all, if i want to realize the function of *showing all the content in the tab* so i must set `[nzShowPagination]="false"` in nz-tabset
2. But i do not need the `prev` and `next` button, so i set `@tabs-scrolling-size: 0px` in `theme.less`
3. But the caculator of the `transform:translate3d()` has bug
4. I think the reason may be in the `line 280 in the nz-tabs-nav.component.ts`, it is hard coding the `PAGINATION_PIX = 64`

vthinkxie:
Hi @NearZXH 
do you have the interest to submit a pr to fix this?

NearZXH:
OK, i will have a try. This will be my first PR.


------------------ 原始邮件 ------------------
发件人: "vthinkxie"<notifications@github.com>;
发送时间: 2019年6月25日(星期二) 下午4:39
收件人: "NG-ZORRO/ng-zorro-antd"<ng-zorro-antd@noreply.github.com>;
抄送: "张晓华"<270761322@qq.com>;"Mention"<mention@noreply.github.com>;
主题: Re: [NG-ZORRO/ng-zorro-antd] nz-tabset click the lastest tab alwayshas 64px margin-right (#3539)




Hi @NearZXH
 do you have the interest to submit a pr to fix this?

PR #3709, opened 2019-07-02
fix(module:tabs): fix the pagnation padding-right when scrolling (#3539)

NearZXH:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #3539 

## What is the new behavior?
will have no 64 padding-right when setting `@tabs-scrolling-size` in theme.less

```
[ ] Yes
[x] No
```


vthinkxie:
good job @NearZXH 
plz add the test and we will review your pr

NearZXH:
@vthinkxie I have add two unit test. I add  function in nz-tabs-nav called `pxToNumber()` which is similar with the `pxToNumber()` in `components/core/util/text-measure.ts`, so i export the function, i'm not sure it's make sense.

components/core/util/text-measure.ts
```diff
@@ -25,7 +25,7 @@ const wrapperStyle = {
   lineHeight: 'inherit'
 };
 
-function pxToNumber(value: string | null): number {
+export function pxToNumber(value: string | null): number {
```
vthinkxie: should use toNumber in `core/util/convert`

components/tabs/nz-tabs-nav.component.ts
```diff
@@ -286,6 +287,19 @@ export class NzTabsNavComponent implements AfterContentChecked, AfterContentInit
     }
   }
 
+  get navContainerScrollPaddingPix(): number {
+    const navContainer = this.navContainerElement.nativeElement;
+    // tslint:disable: no-any
+    const originStyle: CSSStyleDeclaration = window.getComputedStyle
```
vthinkxie: should make sure `this.platform.isBrowser` before use `window`

NearZXH:
@vthinkxie I tried `toNumber()` but i think it's not suitable to such as. the originStyle.paddingLeft is a string such as '100px'，the return value is 0 through `toNumber()`; here is the original code of toNumber():
``` javascript
// in angular cdk: ~/node_modules/@angular/cdk/bundles/cdk-coercion.umd.js:Line:48
/**
 * Whether the provided value is considered a number.
 * \@docs-private
 * @param {?} value
 * @return {?}
 */
function _isNumberValue(value) {
    // parseFloat(value) handles most of the cases we're interested in (it treats null, empty string,
    // and other non-number values as NaN, where Number just uses 0) but it considers the string
    // '123hello' to be a valid number. Therefore we also check if Number(value) is NaN.
    return !isNaN(parseFloat((/** @type {?} */ (value)))) && !isNaN(Number(value));
}
```

vthinkxie:
> @vthinkxie I tried `toNumber()` but i think it's not suitable to such situation. the originStyle.paddingLeft is a string such as '100px'，the return value is 0 through `toNumber()`; here is the original code of toNumber():
> 
> ```js
> // in angular cdk: ~/node_modules/@angular/cdk/bundles/cdk-coercion.umd.js:Line:48
> /**
>  * Whether the provided value is considered a number.
>  * \@docs-private
>  * @param {?} value
>  * @return {?}
>  */
> function _isNumberValue(value) {
>     // parseFloat(value) handles most of the cases we're interested in (it treats null, empty string,
>     // and other non-number values as NaN, where Number just uses 0) but it considers the string
>     // '123hello' to be a valid number. Therefore we also check if Number(value) is NaN.
>     return !isNaN(parseFloat((/** @type {?} */ (value)))) && !isNaN(Number(value));
> }
> ```

you are right, I have resolved the comment

components/tabs/nz-tabs-nav.component.ts
```diff
@@ -286,6 +289,23 @@ export class NzTabsNavComponent implements AfterContentChecked, AfterContentInit
     }
   }
 
+  get navContainerScrollPaddingPix(): number {
+    if (this.platform.isBrowser) {
```
vthinkxie: miss `PlatformModule` in the module

NearZXH: I'm confused about whether it's necessary to add `PlatformModule` in `ngModule`, i found that there is no `PlatformModule` in `NzAffix` and `NzCarouselModule`. Can you give me more reasons to do this? `Platform` is a `Service` which is `@Injectable({providedIn: 'root'})`。 Here is the link to [PlatformModule](https://github.com/angular/components/blob/master/src/cdk/platform/platform-module.ts) and [Platform](https://github.com/angular/components/blob/master/src/cdk/platform/platform.ts) in angular.

vthinkxie: it should be a bug in `NzAffixModule` and `NzCarouselModule` then, we have no idea what changes will happen in `PlatformModule` in the furture.

NearZXH:
I import `PlatformModule` to `NzTabModule`. And I found there are 15 modules which are using `platform` without import `PlatformModule`, do i need to make a new pr to fix it?

vthinkxie:
@NearZXH yes, plz help us fix it, thanks a lot!

Issue #3571, opened 2019-06-14
NzTableComponent should support CDK Datasource


lppedd:
## What problem does this feature solve?
`NzTableComponent` will be able to support virtual scrolling of unknown size data sources.
Currently, unlike `NzListComponent`, the table component supports only a static array as input, even for virtual scrolling. This limits the table capabilities, as it uses the Angular CDK virtual scroll viewport, which supports a `Datasource` implementation.
## What does the proposed API look like?
`NzTableComponent` should accept `nzData: T[] | Datasource<T>;`.

lppedd:
If you haven't got time for this, I can consider a PR, although I can't guarantee on the expected delivery time.

vthinkxie:
Hi @lppedd 
The idea sounds great. Does this `Datasource<T>` support only virtual scroll or both virtual scroll and normal mode?

lppedd:
@vthinkxie The `Datasource` should support only virtual scrolling by default AFAIK.
Its API does state it accepts a `CollectionViewer` ([source](https://github.com/angular/components/blob/master/src/cdk/collections/collection-viewer.ts)), which is implemented and injected by the `cdkVirtualFor` directive.

As you can see `CollectionViewer` does only have a `viewChange: Observable<ListRange>;` field.
Thus, a custom implementation could be created to support a fixed size array, in conjuction with a custom, hypothetical, `ArrayDatasource`, maintained internally the `NzTableComponent`. This will prevent having to modify the public table API.

vthinkxie:
@lppedd I got it, just feel free to submit PR for this, we will review it, thanks a lot for your proposal

lppedd:
@vthinkxie do you prefer maintaining the array/datasource as two separate cases, or do you feel like an homogeneous implementation would be better?

vthinkxie:
I think both solutions are ok, but it has to stay compatible with previous version

lppedd:
@vthinkxie thanks! I will open the PR shortly.

lppedd:
I'm now working on this full-time at work.  
I'm battling with, apparently, a change-detection issue, caused by the multiple templates nesting.
The view is constantly re-rendered, causing an odd infinite-automatic-scroll behavior.

The only solution seems to be `detach`ing from change-detection and `detectChanges` manually on scrolling, e.g.

```
this.cdkVirtualScrollViewport
  .elementScrolled()
[...]

PR #3766, opened 2019-07-10
feat(module:table): handle CDK DataSource for virtual scroll

lppedd:
Opening this PR for discussion only at the moment.  
This is the minimal implementation to support the use of a `DataSource`.  
I need this functionality for a product I'm developing and this would be sufficient at the moment.
I'm especially worried about the problem described in [this](https://github.com/NG-ZORRO/ng-zorro-antd/issues/3571#issuecomment-510059518) comment.

What is missing
- [x] Skeleton while loading data.
 Example:
Peek 11-07-2019 16-34
- [ ] <s>Warning when using incompatible inputs (e.g. `DataSource` and non-virtual scroll)</s>

--------

As a next step, I'd like to leverage the `DataSource` itself, to produce a uniform implementation.  
This means if a user passes in `[nzData]="[...]"`, I'll wrap it around an `ArrayDataSource`.  
This would mean an overall cleaner implementation, maybe delegating some functionalities to a common, base `DataSource`.

Closes https://github.com/NG-ZORRO/ng-zorro-antd/issues/3571

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[X] No
```


components/table/style/index.less
```diff
@@ -6,6 +6,7 @@
 @table-header-sort-active-bg: darken(@table-header-bg, 3%);
 @table-header-filter-active-bg: darken(@table-header-sort-active-bg, 5%);
 @table-selection-column-width: 60px;
+@table-skeleton-to-color: shade(@skeleton-color, 5%);
```
vthinkxie: should not change style here, because all style file are synced with antd react version via robot.

lppedd:
@vthinkxie Could you give a look at the `NzTableComponent` implementation?
Specifically at the `handleDataChange` method and the used classes. What do you think? This is temporary but it is a "road" a like.

components/table/style/index.less
lppedd: @vthinkxie where should I do the edits?

lppedd:
A quick update.  
I have tried many different ways to maintain backward-compatibility without going crazy, but this seems to be the only reasonable one.
I have re-organized/separated files so it's more clear what I have done.

components/table/style/index.less
vthinkxie: It would better to let users define the loading template, it could be empty by default, and we can provide a demo with the skeleton style in our docs

lppedd: @vthinkxie I think this can be a problem. Empty `tr`s do not have height, that's why I used
```
div:before {
  content: '.';
  visibility: hidden;
}
```
I don't know if it will mess up the virtual viewport. Let me try.

components/table/nz-table.component.ts
```diff
@@ -307,4 +333,15 @@ export class NzTableComponent<T = any> implements OnInit, AfterViewInit, OnDestr
     this.destroy$.next();
     this.destroy$.complete();
   }
+
+  private handleDataChange(nzData: NzDataSource<T> | T[]): void {
+    this.data = isDataSource(nzData) ? nzData : new NzArrayDataSource(nzData, !this.nzFrontPagination);
```
vthinkxie: nzData maybe null or undefined

lppedd: @vthinkxie Fixed, if `null | undefined`, an empty array will be used.

components/table/style/index.less
lppedd: @vthinkxie Okay, I pushed a version with your request. Let me know.
However I user must remember to use 
```
content: '.';
visibility: hidden;
```
Or some other trick to maintain the height of the `tr`.

vthinkxie: we could put these styles in the patch.less file if must

lppedd: @vthinkxie I could write something like 
```
tr.loading > * {
  content: '.';
  visibility: hidden;
}
```

So that it matches every first level element put in by the user.
However this seems like "guessing" or "hoping for the best", if you know what I mean.

Do we really want to let users customize the template? It's not really a skeleton like the one used in the `NzListComponent`.

vthinkxie: the key point of this is not providing the feature for users to customize the template, but the style is not from ant design, which means that it may break in a patch or minimal version after sync with antd.

let's put it in the patch.less for temp solution, since there is no better idea to handle the loading data display now.

lppedd: @vthinkxie Okay. I'll revert to internal-only and move styles to `patch.less`.
What about a directive-like Component, which attached to `ŧr`s ?
Styles could be encapsulated there.

vthinkxie: a directive-like Component is a better idea

lppedd: @vthinkxie what about just placing them in the `NzTableComponent` styles? It's sufficient.

lppedd: @vthinkxie done. Also updated english documentation.

lppedd:
I've introduced a check which should help catching "wrong" user-side behavior.
If someone wants to submit an `NzDataSource` implementation with normal mode (non-virtual), the `NzDataSource` must provide a valid `Iterator`.

https://github.com/NG-ZORRO/ng-zorro-antd/blob/246a67875f7a7b6293396b2cfd65aa3831ff046d/components/table/nz-table.component.ts#L343-L347

lppedd:
Just wanted to let you know I haven't stopped working on this.
I'm just waiting for feedback or ideas before writing tests :smile:

components/table/nz-table.component.less
```diff
@@ -0,0 +1,27 @@
+@import '../style/themes/index';
```
vthinkxie: let's put all the style into patch.less in component/table/style

vthinkxie:
everything looks good to me, @wendzhue @hsuanxyz plz take a look to this
thanks a lot to @lppedd you did a great job!

lppedd:
@vthinkxie thanks.
I think we need to come up with an idea to better solve the following use-case:
- DataSource
- front pagination
- non-virtual

Which means
```
<nz-table #table [nzData]="dataSource" [nzFrontPagination]="true" [nzShowPagination]="true">
   ...
   <tr *ngFor="let data of table.data; index as i">
      ...
   </tr>
```

This is particularly tricky.

The ideal workflow would be:
1. user asks for a page, let's say page 2.
2. DataSource is notified, does an HTTP calls, update the internal array
3. change detection runs and calls `[Symbol.iterator]`
4. the table view is updated

However what actually happens is
1. user asks for a page, let's say page 2.
2. following the button click, change detection runs (thanks Zone.js!) and calls `[Symbol.iterator]`
3. table view is updated
3. DataSource is notified, does an HTTP calls, update the internal array

You can clearly see there is a problem, the table is "behind" of an HTTP call, and the data displayed is wrong.

Possible solutions:
1. User inject `ChangeDetectorRef` inside its own `NzDataSource`
```
private fetchRange(range: ListRange): void {
   this.http
       .get(`...`)
       .subscribe(result => {
         this.data = result;
         this.cdr.detectChanges(); // Here
       });
}
```
Two change detection cycles are run then.

2. We wrap the pagination button click in a non-Zone.js area, and somehow *we* invoke change detection.

3. Any other idea? 😄

components/table/nz-table.component.less
lppedd: Done.

components/table/nz-table.component.ts
```diff
@@ -307,4 +322,26 @@ export class NzTableComponent<T = any> implements OnInit, AfterViewInit, OnDestr
...
+
+    if (this.nzFrontPagination) {
+      this.data = new NzDetachedDataSource(this, this.data);
+      this.data.connect(this).subscribe({
+        next: data => this.nzCurrentPageDataChange.emit(data as T[])
+      });
+    }
+  }
+
+  private isIteratorOverriddenOrThrow(nzData: NzDataSource<T>): void {
```
lppedd: This should return a `boolean`, and should be called after `isDataSource(nzData)`

components/table/nz-table.component.ts
```diff
@@ -307,4 +322,26 @@ export class NzTableComponent<T = any> implements OnInit, AfterViewInit, OnDestr
     this.destroy$.next();
     this.destroy$.complete();
   }
+
+  private handleDataChange(nzData?: T[] | NzDataSource<T>): void {
+    if (isDataSource(nzData)) {
```
lppedd: See comment below. This can then become a shorthand `if`

components/table/nz-table.component.ts
```diff
@@ -213,21 +220,24 @@ export class NzTableComponent<T = any> implements OnInit, AfterViewInit, OnDestr
   }
 
   updateFrontPaginationDataIfNeeded(isPageSizeOrDataChange: boolean = false): void {
-    let data = this.nzData || [];
     if (this.nzFrontPagination) {
-      this.nzTotal = data.length;
+      this.nzTotal = this.data.length;
```
lppedd: This is not ideal, as the NzDataSource instance might not have the value available at this point. I need to refactor this, somehow, into an "asynchronous" process

components/table/nz-table.component.ts
```diff
@@ -307,4 +322,26 @@ export class NzTableComponent<T = any> implements OnInit, AfterViewInit, OnDestr
...
+    if (isDataSource(nzData)) {
+      this.isIteratorOverriddenOrThrow(nzData);
+      this.data = nzData;
+    } else {
+      this.data = new NzArrayDataSource(nzData || [], !this.nzFrontPagination);
+    }
+
+    if (this.nzFrontPagination) {
+      this.data = new NzDetachedDataSource(this, this.data);
+      this.data.connect(this).subscribe({
```
lppedd: Better off using (this.data as NzDetachedDataSource<T>).connect()...
It's more clear, as it doesn't make sense to pass this

vthinkxie:
Hi @lppedd 
sorry for the late reply
I'm a little confused about the http calls with `[nzFrontPagination]="true"`

When the nzFrontPagination is true, there should no http calls I think, because all data are loaded at the very beginning.

lppedd:
@vthinkxie but how do you enforce the user to provide a `DataSource` which simply contains an `Array`? I mean, as far as I understood, the difference between `frontPagination=true` and `...=false`, is that when `true`, the `NzTableComponent` handle the data splitting, while with `false`, events signaling pages and sizes are emitted, and the user must handle them himself.

The same applies with the `DataSource`, but it still can have HTTP calls.
Anyway, wouldn't it be the same with a `DataSource` containing only an `Array`? It's still asynchronous by using Observables (sort of).

But maybe I misunderstood what you're trying to say.

vthinkxie:
@lppedd 
I mean if the user set frontPagination = true, why he still needs to send http calls when clicking pagination?

lppedd:
@lppedd well the user isn't forced to do HTTP calls, but it still is an option. We don't have control over the implementation of the inputted `DataSource`.

vthinkxie:
@lppedd 
yes, frontPagination = true should work with virtual scroll if user wants to send http calls, right?
if so I think we can add a note in the document, to tell the user not using frontPagination = true with non-virtual-scroll and I think most users would not use it in most scenarios

lppedd:
@vthinkxie yes absolutely. `DataSource` + `frontPagination = true` + virtual scroll, works nicely always.

Just a question: as I had wrote in the long message above, why not deprecating the use of the internal field `data` in favor of an event, such as `nzCurrentPageDataChange`.  
This doesn't solve the problem, but at least is a step forward.

Anyway, I'll review the code again, and report back if I find something else worth discussing (I don't think so). Then I'll write some tests.

vthinkxie:
@lppedd 
yes, we should move a step forward, but I think most current user works fine with Array input, but I let's talk about the breaking change when 9.0 comes.

lppedd:
@vthinkxie sure, agree. I'll keep you updated also on https://github.com/NG-ZORRO/ng-zorro-antd/pull/3766#discussion_r308003039, which needs some more thinking. If you want to throw ideas of any kind, they're very welcome.

components/table/nz-table.component.ts
vthinkxie: agree, let's do this :)

lppedd:
@vthinkxie Sorry I've been pretty busy with other work. Just wanted to let you know I haven't forgotten.

vthinkxie:
Hi @lppedd 
the table component are refactored in v9 now, and all data process are moved to table-data.service.ts now, it should be much clearer than before
feel free to ping me if you need any help

vthinkxie:
Hi
This Pull Request is going to be closed since no update for a long time, feel free to open a new PR if any new update.

Issue #3530, opened 2019-06-04
Integrate date-picker calendar input auto-focus option and functionality


cnotv:
## What problem does this feature solve?
Allow to optionally automatically focus the date input field on panel open.

This would allow quick editing.

The proposed implementation uses "document.querySelector" which probably you would like to avoid.
## What does the proposed API look like?
```js
@Input() @InputBoolean() calInputAutoFocus: boolean;
@ViewChild('datepicker') datepicker: NzDatePickerComponent;

ngOnInit() {
    if (calInputAutoFocus) {
      this.subscribeInputFocus();
    }

}

private subscribeInputFocus(): Subscription {
    return this.datepicker.nzOnOpenChange.subscribe((open: boolean) => {
      if (open) {
        setTimeout(() => {
          const calInput = document.querySelector(
            'calendar-input input.ant-calendar-input'
          ) as HTMLInputElement;
          calInput.focus();
        });
      }
    });
}
```

cicero-mt:
This would be great!

PR #3146, opened 2019-03-24
fix(module:date-picker): make keyboard navigation possible

jimmytheneutrino:
Tabbing away from date-picker overlay was not possible: It wrapped
around and usually ended up in browser's address bar. Also, once you
selected date-picker wrapper span, there was no way to open the overlay
via keyboard.
This fixes this. Tabbing away from overlay now focuses on date-picker
wrapper span. Pressing enter on the wrapper span opens the overlay.

Please also see a similar PR for the time-picker module #3145 .

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Application (the showcase website) / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Tabbing away from date-picker overlay was not possible: It wrapped
around and usually ended up in browser's address bar. Also, once you
selected date-picker wrapper span, there was no way to open the overlay
via keyboard.

Issue Number: N/A

## What is the new behavior?
This fixes this. Tabbing away from overlay now focuses on date-picker
wrapper span. Pressing enter on the wrapper span opens the overlay.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
The behavior (both current and new) can be observed on the main demo page (npm run site:start).

PR #3804, opened 2019-07-16
fix(module:date-picker): open on enter and focus on inner input

jimmytheneutrino:
Extracted the good parts suitable for a11y from #3146. See discussion there.

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Opening date-picker with keyboard is not possible.

Opening date-picker by a click, does not focus on the inner input element.

Issue Number: N/A

## What is the new behavior?
You can open date-picker pressing enter on the keyboard when the wrapper element is selected.

Once date-picker is open, the inner input element is selected, so it is immediately possible to input date from keyboard. (Of course, this is only applicable to the cases, when there is an inner input element: e.g., month and year pickers do not have any).

```
[ ] Yes
[X] No
```

### DIFF
--- a/components/date-picker/lib/popups/date-range-popup.component.html
+++ b/components/date-picker/lib/popups/date-range-popup.component.html
@@ -33,6 +33,7 @@
     [locale]="locale"
     [disabledDate]="disabledDate"
     [format]="format"
+    [autoFocus]="partType !== 'right'"
     [placeholder]="getPlaceholder(partType)"
   ></calendar-input>
 </ng-template>
--- a/components/date-picker/picker.component.ts
+++ b/components/date-picker/picker.component.ts
@@ -105,14 +105,18 @@ export class NzPickerComponent implements OnInit, AfterViewInit {
 
   ngAfterViewInit(): void {
     if (this.autoFocus) {
-      if (this.isRange) {
-        const firstInput = (this.pickerInput.nativeElement as HTMLElement).querySelector(
-          'input:first-child'
-        ) as HTMLInputElement;
-        firstInput.focus(); // Focus on the first input
-      } else {
-        this.pickerInput.nativeElement.focus();
-      }
+      this.focus();
+    }
+  }
+
+  focus(): void {
+    if (this.isRange) {
+      const firstInput = (this.pickerInput.nativeElement as HTMLElement).querySelector(
+        'input:first-child'
+      ) as HTMLInputElement;
+      firstInput.focus(); // Focus on the first input
+    } else {
+      this.pickerInput.nativeElement.focus();
     }
   }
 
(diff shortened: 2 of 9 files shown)


components/date-picker/picker.component.ts
```diff
@@ -135,6 +139,7 @@ export class NzPickerComponent implements OnInit, AfterViewInit {
     if (this.realOpenState) {
       this.overlayOpen = false;
       this.openChange.emit(this.overlayOpen);
+      this.focus();
```
wenqi73: Need a test for `focus`.

components/date-picker/lib/popups/date-range-popup.component.html
```diff
@@ -33,6 +33,7 @@
     [locale]="locale"
     [disabledDate]="disabledDate"
     [format]="format"
+    [autoFocus]="partType != 'right'"
```
wenqi73:
```suggestion
    [autoFocus]="partType !== 'right'"
```

wenqi73:
Thank you @jimmytheneutrino, this will close #3530.

components/date-picker/picker.component.ts
jimmytheneutrino: Added 'should focus on the trigger after a click outside' tests

components/date-picker/lib/popups/date-range-popup.component.html
jimmytheneutrino: fixed

jimmytheneutrino:
@wendzhue 
Of these 2 problems, 2 seemed to be easy to do and quite useful, so I included a fix and a test.

However, 1 seems to be more subtle. A probable fix (which I am not really aware of) requiring an enter on changes could possibly disrupt the behavior of different components, because calendar-input is used in many places. Besides, it is not really an improvement but conformance to the react version, which is not necessarily 100% better. Lastly, it is not really that connected to the PR at hand. I think it is better to state 1 in a separate issue.

jimmytheneutrino:
@wendzhue 
After the fix for 2, the codecov patch check failed no matter what. So it seemed easier to try a fix for 1 that would also probably eliminate codecov issues. Fixed the tests accordingly.

jimmytheneutrino:
It seems CI has failed with some strange 'Chromium disconnected' error. Rerun?

jimmytheneutrino:
The test job does not seem to be able to finish. (It is ok on my local machine.)
Please see, e.g., this thread for a similar mistake: https://github.com/karma-runner/karma-chrome-launcher/issues/196

components/date-picker/picker.component.ts
haven110: image
is this function implemented? The pop-up date box automatically captures the focus inside and can be entered by the keyboard

PR #3878, opened 2019-07-29
feat(site): add two ways to compile target demo when running start

orzyyyy:
## PR Type

Imporve compiling site.

## How to use

Suppose the demo you want to compile is `button/demo/ghost.md`, then

1. run command `npm run start button/demo/ghost.md`

2. add `only: true` at the top, which is called `metadata`, as below,
```bash
---
only: true
order: 8
title:
  zh-CN: 幽灵按钮
  en-US: Ghost Button
---
```

### DIFF
--- a/components/affix/demo/basic.md
+++ b/components/affix/demo/basic.md
@@ -1,4 +1,5 @@
 ---
+only: true
 order: 0
 title:
   zh-CN: 基本
--- a/components/affix/demo/basic.md
+++ b/components/affix/demo/basic.md
@@ -1,5 +1,4 @@
 ---
-only: true
 order: 0
 title:
   zh-CN: 基本
--- a/components/button/demo/block.md
+++ b/components/button/demo/block.md
@@ -1,4 +1,5 @@
 ---
+only: true
 order: 9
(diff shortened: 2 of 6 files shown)


orzyyyy:
深度截图_选择区域_20190729122445

orzyyyy:
Any progress?

wendzhue:
It's not a good idea to write

``` yaml
only: true
```

Better to make a parameter of command `npm run start`, like `npm run start button`.

orzyyyy:
CI won't pass if any commit contains `only`, so I think it doesn't matter.

wendzhue:
> CI won't pass if any commit contains `only`, so I think it doesn't matter.

CI takes time. And developers still needs to add a commit for remove that line.

orzyyyy:
OK, I'll add a check before commit.

orzyyyy:
Now you can run `npm start button/demo/ghost.md` to compile that one.

orzyyyy:
Ping~

wendzhue:
@hsuanxyz Have time to review this?

PR #4036, opened 2019-08-21
revert: table small sticky style

wenqi73:
Reverts NG-ZORRO/ng-zorro-antd#3849
It will cause another issue of small table hover style. Ant design has fixed small sticky style (https://github.com/NG-ZORRO/ng-zorro-antd/commit/0608f4596bce89ad5b59124adfdecf182e8038b5#diff-6f1e26fcf688fcefd89a8725673c03abR78)

### DIFF
--- a/components/table/style/patch.less
+++ b/components/table/style/patch.less
@@ -153,12 +153,6 @@
           background: @component-background;
         }
       }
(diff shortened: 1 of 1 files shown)


vthinkxie:
should add hover style, not just remove them, the sticky is not the same as react version.

PR #5176, opened 2020-04-29
fix(module:table): hiding table lose column width

### LINKED ISSUE #5160: Show table again after hiding table lose column width
### Reproduction link
[https://stackblitz.com/edit/angular-hgjmxe](https://stackblitz.com/edit/angular-hgjmxe)

### Steps to reproduce
click show/hidden button, Table column width is not the initial width

### What is expected?
Table column width maintain initial width

Table columns are not the width initially set

| Environment | Info |
|---|---|
| ng-zorro-antd | 9.1.0 |
| Browser | Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.132 Safari/537.36 |

zhmushan:
Fix #5160

### DIFF
--- a/components/table/src/table-style.service.ts
+++ b/components/table/src/table-style.service.ts
@@ -30,9 +30,8 @@ export class NzTableStyleService {
   listOfListOfThWidthPx$ = merge(
     this.manualWidthConfigPx$,
     combineLatest([this.listOfAutoWidthPx$, this.manualWidthConfigPx$]).pipe(
-      map(([autoWidth, manualWidth]) => {
-        /** use autoWidth until column length match **/
-        return autoWidth.length !== manualWidth.length ? manualWidth : autoWidth;
+      map(([_, manualWidth]) => {
+        return manualWidth;
       })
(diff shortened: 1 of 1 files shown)


zhmushan:
@vthinkxie thanks for your reply, I didn't understand the meaning of this line before. I will check it later.

zhmushan:
@vthinkxie I have tried many situations, it seems that I can simply return `manualWidth` in any situation. I'm not sure if I missed something

vthinkxie:
this would totally break the autoWidth calculation feature, there is no need to pass the value to the nzWidth in v9, all width are auto calulated

zhmushan:
@vthinkxie Sorry, I don’t seem to be able to solve this issue. I temporarily use `ngif` to hide the form in my project. I will pay attention to future updates.

vthinkxie:
Hi @zhmushan thanks for your PR anyway!

PR #5375, opened 2020-06-03
feat(module:select): open the select when tabbing to it in a form

kingbreddo:
When tabbing inside a form it was not clear which select was focused.
By automatically opening the select the user sees which form is now focused.

## PR Checklist

- [N/A] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When tabbing in a form the select is not mark as focused (So a blue border around it)

Issue Number: N/A

## What is the new behavior?
The select will be opened automatticly when tabbing to it.

```
[ ] Yes
[x] No
```


components/select/select.component.ts
```diff
@@ -352,6 +353,16 @@ export class NzSelectComponent implements ControlValueAccessor, OnInit, AfterVie
     this.clearInput();
   }
 
+  onKeyUp(e: KeyboardEvent): void {
```
vthinkxie, line 356: is there any special reason to add this into the `(keyup)` function?
it would have conflict with onKeyDown and TAB keyCode

kingbreddo: Yes when your focus is on a text field and you tab to the next select field. It will trigger the keyUp on the text field and a keyDown on the select.

To be honest I don't really know what the usecase is of the TAB usage inside the keyDown.

vthinkxie: it works when pressing TAB when the user has already opened the select panel

kingbreddo: Sorry I don't understand your last comment.

But i've tested if it will work with `setOpenState(true)` inside the `onKeyDown` but it won't.
This code really should live inside the onKeyUp otherwise it will not work.

When I manually test it with the documentation site it opens the select when tabbing to it (possible through onKeyUp). And it closes it nicely after tabbing away from it E.g. to the submit button (possible through onKeyDown).

You can test it locally by using the following form inside the documentation website: http://localhost:58046/components/form/en#components-form-demo-coordinated

If I need something to change feel free to say it :)

kingbreddo: @vthinkxie Any update / suggestion on this code change request?

ousc:
For this, what new progress.

PR #5939, opened 2020-10-16
fix(doc layout): duplicated binding description

samber:
nzCollapsed is describe twice

https://ng.ant.design/components/layout/en

image


vthinkxie (approved):
lgtm

PR #6044, opened 2020-11-11
fix(module:code-editor): memorize cursor position and selections

### LINKED ISSUE #6038: code-editor jumps mouse cursor when model changes
### Reproduction link

Cannot be reproduced on StackBlitz because nz-code-editor doesn't work there.

### Steps to reproduce

Use `nz-code-editor` as a form control (ngModel, formControlName) and trigger a model→view update (e.g. update ngModel) while the cursor is somewhere in the middle of the document.

### What is expected?

The cursor position should be retained.

The cursor jumps to the beginning of the document.

This is actually intended behavior in monaco, also see https://github.com/Microsoft/monaco-editor/issues/1397 or [CursorChangeReason#ContentFlush](https://microsoft.github.io/monaco-editor/api/enums/monaco.editor.cursorchangereason.html). I believe the fix is for nz-code-editor's ControlValueAccessor not to call setValue, but execute an edit instead.

| Environment | Info |
|---|---|
| ng-zorro-antd | 10.1.1 |
| Browser | Chrome |

Airblader:
Calling #setValue() on a monaco editor instance causes the mouse cursor to jump to the
beginning of the document, which is undesirable for a ControlValueAccessor implementation
which should only update the content. Also see
https://github.com/microsoft/monaco-editor/issues/1397 for more information.

Unfortunately, using #executeEdits() doesn't really work as a replacement because it will
mess with selections. Therefore we memorize and restore the cursor position before/after
setting the value. Since #setValue() also causes a selection to be lost we do the same for
those. Restoring selections may produce slightly unexpected results if the selection covers
part of the content that has been changed, but is still preferable to losing the selection
altogether.

fixes #6038

## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features) (there are no working code-editor tests)
- [ ] Docs have been added / updated (for bug fixes / features) (not required)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/code-editor/code-editor.component.ts
+++ b/components/code-editor/code-editor.component.ts
@@ -212,7 +212,8 @@ export class NzCodeEditorComponent implements OnDestroy, AfterViewInit {
 
     if (this.nzEditorMode === 'normal') {
       if (this.modelSet) {
-        (this.editorInstance.getModel() as ITextModel).setValue(this.value);
+        const model = this.editorInstance.getModel() as ITextModel;
+        this.preservePositionAndSelections(() => model.setValue(this.value));
       } else {
         (this.editorInstance as IStandaloneCodeEditor).setModel(
           monaco.editor.createModel(this.value, (this.editorOptionCached as EditorOptions).language)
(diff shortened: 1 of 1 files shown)


Airblader:
CC @sorahn

hsuanxyz (approved):

Issue #5907, opened 2020-10-13
[Input] text-area support value count


hsuanxyz:
https://github.com/ant-design/ant-design/pull/26952
![image](https://user-images.githubusercontent.com/22736418/95812396-e0c86280-0d47-11eb-98a0-c65e258191c2.png)

PR #6104, opened 2020-11-26
feat(module:input): textarea supports character count

yangjunhan:
close NG-ZORRO#5907

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #5907 

## What is the new behavior?
It now supports displaying character count at the bottom-right corner below the textarea. User can customize max length and character count computation function.

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/input/textarea-count.component.ts
+++ b/components/input/textarea-count.component.ts
@@ -1,0 +1,76 @@
+/**
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
+ */
+import {
+  AfterContentInit,
+  ChangeDetectionStrategy,
+  Component,
+  ContentChild,
+  ElementRef,
+  Input,
+  isDevMode,
+  OnDestroy,
+  Renderer2
+} from '@angular/core';
+
+import { EMPTY, merge, Subject } from 'rxjs';
+import { map, startWith, takeUntil } from 'rxjs/operators';
+
+import { isNotNil } from 'ng-zorro-antd/core/util';
+
+import { NzInputDirective } from './input.directive';
+
+@Component({
+  selector: 'nz-textarea-count',
+  template: `
+    <ng-content select="textarea[nz-input]"></ng-content>
+  `,
+  host: {
+    class: 'ant-input-textarea-show-count'
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class NzTextareaCountComponent implements AfterContentInit, OnDestroy {
+  @ContentChild(NzInputDirective, { static: true }) nzInputDirective!: NzInputDirective;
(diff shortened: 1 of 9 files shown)


components/input/textarea-count.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  template: `
+    <ng-content select="textarea[nz-input]"></ng-content>
+  `,
+  host: {
+    class: 'ant-input-textarea-show-count'
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class NzTextareaCountComponent implements AfterContentInit, OnDestroy {
+  @ContentChild(NzInputDirective, { static: true }) nzInputDirective!: NzInputDirective;
```
Airblader, line 22: This doesn't need to be static: true, does it?

components/input/textarea-count.component.ts
```diff
@@ -0,0 +1,54 @@
...
+          map(() => this.nzInputDirective.ngControl.value),
+          startWith(this.nzInputDirective.ngControl.value as string)
+        )
+        .subscribe(value => {
+          this.onValueChange(value);
+        });
+    }
+  }
+
+  onValueChange(value: unknown): void {
```
Airblader, line 44: I don't think this needs to be unknown, the value of a textarea is always a string (and above there's also a type assertion for it, too). Also makes `String(value)` superfluous.

components/input/textarea-count.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    <ng-content select="textarea[nz-input]"></ng-content>
+  `,
+  host: {
+    class: 'ant-input-textarea-show-count'
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class NzTextareaCountComponent implements AfterContentInit, OnDestroy {
+  @ContentChild(NzInputDirective, { static: true }) nzInputDirective!: NzInputDirective;
+  @Input() nzMaxCharacterCount: number = 0;
```
Airblader, line 23: Changing either the maximum count or the compute function will not trigger the display to be updated.

components/input/textarea-count.component.ts
stygian-desolator: The value of a textarea could be an emoji, like 😄

Airblader: That'd still be a string.

components/input/textarea-count.component.ts
hsuanxyz: thx review, dynamic content will make matching targets more complex, in which case I think it's enough for developers.

Airblader: The static flag isn't about dynamic content, it's about lifecycle, and static true is deprecated so it has to be removed anyway. In here it can be removed without any additional effort.

PR #6383, opened 2021-01-27
fix(module:dropdown): completes mouseState$ on component destroy

lacroixdavid1:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

memory leak in dropdown

## What is the new behavior?
Subject is now completed to fix the memory leak.

```
[x] Yes
[ ] No
```

### DIFF
--- a/components/dropdown/dropdown-menu.component.ts
+++ b/components/dropdown/dropdown-menu.component.ts
@@ -105,6 +105,7 @@ export class NzDropdownMenuComponent implements AfterContentInit, OnDestroy, OnI
   }
 
   ngOnDestroy(): void {
+    this.mouseState$.complete();
     this.destroy$.next();
     this.destroy$.complete();
   }


vthinkxie (changes requested):
Hi, there won't be any memory leaks here
there is not any subscription that subscribes to mouseState$ and not destroyed onDestroy

lacroixdavid1:
The BehaviorSubject stays open and its current value stays in memory. Every subject instance should be completed when they won't emit any more. Also, by completing the subject, every subscription ends.

vthinkxie:
> The BehaviorSubject stays open and its current value stays in memory. Every subject instance should be completed when they won't emit any more. Also, by completing the subject, every subscription ends.

Hi, every component has its destroy$ to handle the subscription, a subject without any active subscription is a function, there won't be memory leaks AFAIK

Issue #5844, opened 2020-09-29
Datepicker do not close on tab out


valerio-lucantonio:
### Reproduction link
[https://stackblitz.com/edit/waxmwq--run?file=src/app/app.component.ts](https://stackblitz.com/edit/waxmwq--run?file=src/app/app.component.ts)

### Steps to reproduce
1) enter a date in the picker typing it with the keyboard
2) tab out

### What is expected?
the datepicker close after tab out

### What is actually happening?
datepicker keeps open while the focus moves to the next field in the form

| Environment | Info |
|---|---|
| ng-zorro-antd | 10.0.0 |
| Browser | Chrome (but i suppose is the same on every browser) |

valerio-lucantonio:
Hello, any news/acknowledge on this issue?

arneysebaert:
This bug can still be reproduced, any progress/news on it?

PR #6571, opened 2021-04-01
fix(module:date-picker): close date-picker after tabbing out

KeelanS:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: [5844](https://github.com/NG-ZORRO/ng-zorro-antd/issues/5844)

## What is the new behavior?

When tabbing out of the datepicker the picker will close.
When navigating in a range picker, the picker will only close if you go outside of the datepicker. Tabbing between the left and right side of the range picker will not close the picker. 

This works both while tabbing or shift-tabbing.

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/date-picker/date-picker.component.ts
+++ b/components/date-picker/date-picker.component.ts
@@ -107,8 +107,7 @@ export type NzDatePickerSizeType = 'large' | 'default' | 'small';
     '[class.ant-picker-rtl]': `dir === 'rtl'`,
     '[class.ant-picker-borderless]': `nzBorderless`,
     '[class.ant-picker-inline]': `nzInline`,
-    '(click)': 'picker.onClickInputBox($event)',
-    '(keydown)': 'onKeydown($event)'
+    '(click)': 'picker.onClickInputBox($event)'
   },
   providers: [
     DatePickerService,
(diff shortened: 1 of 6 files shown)


KeelanS:
This fix worked out of the box for all examples on the demo page except for the 'start-end' one where you create a custom range picker. Pressing tab while in the left input would open the right picker, but wouldn't select the input itself which meant that you couldn't write your date.

Since it was a custom implementation I adapted it to work with my fix. This did remove the usage of the ngOnOpenChange outputs. If this is a problem I could change it back and try to fix it another way. But I have tried doing that and it's not that obvious. 

Anyway I'd appreciate any input on this :)

wenqi73:
@KeelanS Should not just use `tab` as the judgment. After clicking `tab`, just see the active element is in the input, if not, close the panel, and the demo will not be affected. BTW, add a test for range picker, that will be better.

KeelanS:
@wenqi73 You mean just on the 'start-end' demo component or everywhere?

Right now this component works like I would think it should work. One assumption I made was that after a 'tab', the focus would go on the other input without the picker opening.

I guess this example is still OK since it still shows you can use the 'open' method and the 'nzDisabledDate' to customize your datepickers. And that was the main purpose of this example.

components/date-picker/date-picker.component.ts
```diff
@@ -421,4 +422,19 @@ export class NzDatePickerComponent implements OnInit, OnChanges, OnDestroy, Cont
       }
     }
   }
+
+  onKeydown(event: KeyboardEvent): void {
+    if (!this.isRange && event.key === 'Tab') {
+      this.close();
+    } else if (this.isRange && event.key === 'Tab') {
+      if (this.datePickerService.activeInput === 'left' && event.shiftKey) {
```
wenqi73, line 430: The `tab` action may be changed in the future. I think this way is better: 
```ts
if (!this.elementRef.nativeElement.contains(document.activeElement)) { 
  // ...
}
```

KeelanS: I had to check on focusEvent since at both keyDown and keyUp, the nativeElement and activeElement would be the same.

components/date-picker/demo/start-end.ts
```diff
@@ -5,6 +5,7 @@ import { NzDatePickerComponent } from 'ng-zorro-antd/date-picker';
   selector: 'nz-demo-date-picker-start-end',
   template: `
     <nz-date-picker
+      #startDatePicker
```
wenqi73, line 8: Useless code

wenqi73 (approved):
LGTM

wenqi73:
@KeelanS Thanks!

PR #6602, opened 2021-04-13
fix(module:time-picker): close time-picker after tabbing out

KeelanS:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The timepicker stays open when you tab out of it.

## What is the new behavior?

When tabbing out of the datepicker, the picker will close.

```
[ ] Yes
[x] No
```

## Other information
This uses the same method as my previous pr that fixed the same issue on the datepicker: [datepicker fix](https://github.com/NG-ZORRO/ng-zorro-antd/pull/6571)


components/time-picker/time-picker.component.ts
```diff
@@ -240,8 +240,11 @@ export class NzTimePickerComponent implements ControlValueAccessor, OnInit, Afte
     }
   }
 
-  onFocus(value: boolean): void {
-    this.focused = value;
+  onFocusChange(event: FocusEvent): void {
+    this.focused = event.type === 'focus';
+    if (!this.focused) {
+      this.close();
+    }
```
wenqi73, line 247:
```suggestion
    if (!value) {
      this.checkAndClose();
    }
```
The `checkAndClose` should contains two situations:
- If time is valid should change value and close, ref `onKeyupEnter()`;
- If time is invalid should reset to last value and close, ref `onKeyupEsc()`;

and these should be added to the tests.

KeelanS: Thanks for the feedback.

This validation is actually already done in the onInputChange() method so I think I can get away with just adding setCurrentValueAndClose() that is already being used in the onClickOutside().

There is only one thing I'm not sure about. When you enter an invalid date or nothing, the value will change to the current dateTime. Is this intended behaviour? Because this means that you can't empty the field without clicking the x inside the component.

wenqi73 (changes requested):
Need to determine whether the value is valid or not.

components/time-picker/time-picker.component.ts
```diff
@@ -240,8 +240,11 @@ export class NzTimePickerComponent implements ControlValueAccessor, OnInit, Afte
     }
   }
 
-  onFocus(value: boolean): void {
-    this.focused = value;
+  onFocusChange(event: FocusEvent): void {
```
wenqi73, line 243: No need to use `event` here. It's a `blur` event If it equals `false`.

components/time-picker/time-picker.component.spec.ts
```diff
@@ -191,21 +251,31 @@ describe('time-picker', () => {
...
-      [(ngModel)]="date"
-      (ngModelChange)="onChange($event)"
-      [(nzOpen)]="open"
-      (nzOpenChange)="openChange($event)"
-      [nzDisabled]="disabled"
-      [nzUse12Hours]="use12Hours"
-      [nzSuffixIcon]="nzSuffixIcon"
-      [nzBackdrop]="nzBackdrop"
-    ></nz-time-picker>
+    <ng-container>
```
wenqi73, line 266:
```suggestion
```

wenqi73 (approved):
LGTM. Thanks! @KeelanS

PR #6839, opened 2021-07-11
fix(module:icon): resolve memory leak

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

I've noticed the `NzIconService` cannot be GC'd since it subscribes `configUpdated$` which is a subject and keeps subscribers in its `observers` property until it's teardowned. The creates a memory leak for server-side rendered application since the `NzIconService` is created and destroyed there per each HTTP request.

image

## What is the new behavior?

The `NzIconService` is GC'd successfully.

```
[x] No
```

### DIFF
--- a/components/icon/icon.service.ts
+++ b/components/icon/icon.service.ts
@@ -5,9 +5,9 @@
 
 import { DOCUMENT } from '@angular/common';
 import { HttpBackend } from '@angular/common/http';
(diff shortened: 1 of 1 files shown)


wendellhu95 (approved):
LGTM. Thanks!

PR #6841, opened 2021-07-12
fix(module:drawer): trigger change detection only if there are `nzOnViewInit` listeners

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

The drawer component schedules the DOM timer which causes Angular to run the change detection even if there're no `nzOnViewInit` listeners. This sometimes causes frame drops if the composite thread is busy doing layout stuff but can't delay that timer until next  message loop tick.

## What is the new behavior?

It won't schedule the timer and trigger `tick()` until there're `nzOnViewInit` listeners.

```
[x] No
```


hsuanxyz (approved):
LGTM

PR #6847, opened 2021-07-12
fix(module:code-editor): dispose the event listener when the component is destroyed

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

Monaco's models are stored independently from the editor instance (inside the `ModelService`). So they also have to be disposed separately.

## What is the new behavior?

The event listener is disposed when the component is destroyed.

```
[x] No
```


hsuanxyz (approved):
LGTM

PR #6849, opened 2021-07-12
fix(module:button): do not trigger change detections on click events

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

The button component is probably one of the most used components. Currently, button components trigger change detections on each click event, tho they shouldn't. The click handler only checks if the event target is an anchor and then calls native DOM API, which doesn't require Angular to run `tick()`.

## What is the new behavior?

Button components do not trigger change detection anymore.

```
[x] No
```

### DIFF
--- a/components/button/button.component.ts
+++ b/components/button/button.component.ts
@@ -13,6 +13,7 @@ import {
   ContentChild,
   ElementRef,
   Input,
+  NgZone,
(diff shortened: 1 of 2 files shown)


simplejason (approved):
LGTM

PR #6850, opened 2021-07-12
fix(module:breadcrumb): do not re-enter the Angular zone when calling `navigate`

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

Currently, the `navigate` method re-enters the Angular zone by calling `ngZone.run`. The `navigate` method is a `click` event handler, template event listeners are added within the Angular zone hence there's no need to re-enter it.

Each `ngZone.run` invocation triggers zone's `onInvoke` lifecycle hook which runs `checkStable()` and might trigger the change detection (considering the number of tasks in the current message loop tick).

## What is the new behavior?

The Angular's zone is not re-entered when the `navigate()` is called.

```
[x] No
```

### DIFF
--- a/components/breadcrumb/breadcrumb.component.ts
+++ b/components/breadcrumb/breadcrumb.component.ts
@@ -11,7 +11,6 @@ import {
   ElementRef,
   Injector,
   Input,
-  NgZone,
   OnDestroy,
   OnInit,
   Optional,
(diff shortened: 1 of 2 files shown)


wendellhu95 (approved):
LGTM. Thanks!

PR #6851, opened 2021-07-12
fix(module:autocomplete): resolve memory leak

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

Currently, the `NzAutocompleteComponent` is not GC'd successfully since its subscriptions keep referencing subscribers, subscribers reference `next` callback and all `next` callbacks capture `this`, preventing the `NzAutocompleteComponent` from being GC'd.

This can be observed on the below image. The `__ngContext__` is an `lView`, it's an array. `__ngContext__[2]` is `LViewFlags`, which is a bit mask for the current view state. `256` is `LViewFlags.Destroyed = 0b000100000000`.

By running `(temp1.__ngContext__[2] & 256) === 256` we can see that the component is in destroyed state, but it's still not garbage collected:

image

## What is the new behavior?

The `NzAutocompleteComponent` is GC'd successfully.

```
[x] No
```

### DIFF
--- a/components/auto-complete/autocomplete.component.ts
+++ b/components/auto-complete/autocomplete.component.ts
@@ -131,9 +131,9 @@ export class NzAutocompleteComponent implements AfterContentInit, AfterViewInit,
   @ViewChild('content', { static: false }) content?: ElementRef;
 
   private activeItemIndex: number = -1;
-  private selectionChangeSubscription = Subscription.EMPTY;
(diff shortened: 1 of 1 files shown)


simplejason:
I think `unsubscribe()` will release resources or cancel Observable executions and it can prevent this memory leaks. Do you think we must set any `Subscription` to be null like this @hsuanxyz ?

hsuanxyz (approved):

PR #6852, opened 2021-07-12
fix(module:core): resolve memory leak

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
```

I've noticed the `NzResizeService` cannot be GC'd since it has the `handler` instance property which captures `this`. This creates a memory leak for server-side rendered applications since the `NzResizeService` is created and destroyed there per each HTTP request:

image

## What is the new behavior?

The `NzResizeService` is GC'd successfully.

```
[x] No
```

### DIFF
--- a/components/core/services/resize.ts
+++ b/components/core/services/resize.ts
@@ -3,7 +3,7 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { Injectable, NgZone, Renderer2, RendererFactory2 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


components/core/services/resize.ts
```diff
@@ -31,6 +31,12 @@ export class NzResizeService {
     this.renderer = this.rendererFactory2.createRenderer(null, null);
   }
 
+  ngOnDestroy(): void {
+    // Caretaker note: the `handler` is an instance property (it's not defined on the class prototype).
+    // The `handler` captures `this` and prevents the `NzResizeService` from being GC'd.
+    this.handler = NOOP;
```
wendellhu95, line 37: IMHO, the memory is freed when the `NzResizeService` destroys. Could you please make a reproduction indicating that it do leak here?

PR #6858, opened 2021-07-13
perf(module:select): do not run change detection if the `triggerWidth` has not been changed

arturovt:
## PR Checklist

## PR Type
```
```

Currently, the `nz-select` calls `updateCdkConnectedOverlayStatus` every time the select is opened or closed. The `updateCdkConnectedOverlayStatus` schedules a frame callback which is triggering the whole layout update (due to `getBoundingClientRect()`) and it calls `markForCheck()` which forces Angular to run the change detection from the root component down to `nz-select`.

It causes frame drops since the main thread is busy with:
1) updating layout
2) executing `tick()`

We don't need to run the `markForCheck()` since we don't need the "global" change detection. The `triggerWidth` is bound to `cdkConnectedOverlayMinWidth`, the `triggerWidth` is a number, number bindings are not updated each change detection cycle, since Angular uses `Object.is` comparison to determine if the binding has been changed or not:

```js
function bindingUpdated(lView, bindingIndex, value) {
  const oldValue = lView[bindingIndex];
  if (Object.is(oldValue, value)) {
    return false;
  }
}
```

This means that if the `triggerWidth` hasn't been updated, the `cdkConnectedOverlayMinWidth` binding will not be re-evaluated. This means we have a "dead" change detection (the `tick()` has been run, but there's nothing to update).

image

## What is the new behavior?

The change detection is not running anymore from the root component when the frame callback fires since it'll check if the `triggerWidth` has been changed or not.

```
[x] No
```

### DIFF
--- a/components/select/select.component.ts
+++ b/components/select/select.component.ts
@@ -37,7 +37,7 @@ import { startWith, switchMap, takeUntil } from 'rxjs/operators';
 import { slideMotion } from 'ng-zorro-antd/core/animation';
 import { NzConfigKey, NzConfigService, WithConfig } from 'ng-zorro-antd/core/config';
 import { NzNoAnimationDirective } from 'ng-zorro-antd/core/no-animation';
-import { reqAnimFrame } from 'ng-zorro-antd/core/polyfill';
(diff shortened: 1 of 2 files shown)


hsuanxyz (approved):
LGTM

PR #6982, opened 2021-09-17
fix(module:i18n): update en_GB.ts translations

Bullsized:
Adding missing translations from the en_US file to the en_GB file.

### DIFF
--- a/components/i18n/languages/en_GB.ts
+++ b/components/i18n/languages/en_GB.ts
@@ -20,6 +20,14 @@ export default {
   DatePicker: {
     lang: {
       placeholder: 'Select date',
+      yearPlaceholder: 'Select year',
+      quarterPlaceholder: 'Select quarter',
+      monthPlaceholder: 'Select month',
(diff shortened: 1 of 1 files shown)


components/i18n/languages/en_GB.ts
```diff
@@ -20,6 +20,14 @@ export default {
   DatePicker: {
     lang: {
       placeholder: 'Select date',
+      yearPlaceholder: 'Select year',
+      quarterPlaceholder: 'Select quarter',
+      monthPlaceholder: 'Select month',
+      weekPlaceholder: 'Select week',
+      rangePlaceholder: ['Start date', 'End date'],
```
wenqi73, line 27: Repeated `rangePlaceholder`

wenqi73 (changes requested):

components/i18n/languages/en_GB.ts
Bullsized: this is fixed now, can we merge this?

PR #7093, opened 2021-12-01
perf(module:*): tree-shake warnings in production

arturovt:
## PR Checklist

## PR Type
```
```

```
[ ] Yes
[x] No
```


components/code-editor/code-editor.component.ts
```diff
@@ -214,7 +214,9 @@ export class NzCodeEditorComponent implements OnDestroy, AfterViewInit {
     }
 
     if (this.nzFullControl && this.value) {
-      warn(`should not set value when you are using full control mode! It would result in ambiguous data flow!`);
+      if (typeof ngDevMode === 'undefined' || ngDevMode) {
```
wendellhu95, line 217: Extract a function here?

arturovt: Hey. That's not possible since functions are _runtime_ and `ngDevMode` is a compile time variable.

Terser doesn't call functions during minification & sealing to know their return type (e.g. if it's a static boolean value, like `function doSmth() { return someBoolean }`).

E.g. this is an example of how Material is doing that and they also do duplicate the same condition every time:

image

wendellhu95: So Terser would remove these `if` branches in prodution?

wendellhu95: Ok. I didn't know Terser would do that. Thanks for clarifying.

Issue #7216, opened 2022-01-18
select 选择器bug


vitasyzhi:
### Reproduction link
[https://stackblitz.com/edit/angular-k8bqqb](https://stackblitz.com/edit/angular-k8bqqb)

### Steps to reproduce
在结果load出来前，按 上/下键 进行选择，后台会报错。

core.mjs:6469 ERROR TypeError: Cannot read properties of undefined (reading 'nzValue')
case DOWN_ARROW:
                e.preventDefault();
                if (this.nzOpen) {
                    const nextIndex = activatedIndex < listOfFilteredOptionNotDisabled.length - 1 ? activatedIndex   1 : 0;
                    this.activatedValue = listOfFilteredOptionNotDisabled[nextIndex].nzValue;

### What is expected?
不影响使用

### What is actually happening?
不影响使用

| Environment | Info |
|---|---|
| ng-zorro-antd | 13.0.0 |
| Browser | chrome |

PR #7222, opened 2022-01-19
fix(module:select): fix keyboard event error when option data is empty

chenc041:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: [#7216](https://github.com/NG-ZORRO/ng-zorro-antd/issues/7216)

## What is the new behavior?
Does not do anything, if option data is null

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/select/select.component.ts
+++ b/components/select/select.component.ts
@@ -400,21 +400,17 @@ export class NzSelectComponent implements ControlValueAccessor, OnInit, AfterCon
     const activatedIndex = listOfFilteredOptionNotDisabled.findIndex(item =>
       this.compareWith(item.nzValue, this.activatedValue)
     );
-    // check option value is exist
-    if (!listOfFilteredOptionNotDisabled.length) {
-      return;
-    }
     switch (e.keyCode) {
       case UP_ARROW:
         e.preventDefault();
-        if (this.nzOpen) {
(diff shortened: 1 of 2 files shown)


components/select/select.component.ts
```diff
@@ -400,6 +400,10 @@ export class NzSelectComponent implements ControlValueAccessor, OnInit, AfterCon
     const activatedIndex = listOfFilteredOptionNotDisabled.findIndex(item =>
       this.compareWith(item.nzValue, this.activatedValue)
     );
+    // check option value is exist
+    if (!listOfFilteredOptionNotDisabled.length) {
+      return;
+    }
```
wenqi73, line 406: This logic should not be placed here, it only affects the actions of `UP_ARROW` `DOWN_ARROW` `ENTER`.
We don't have enough tests here, it would be better if you added tests here.

wenqi73 (changes requested):

components/select/select.component.ts
chenc041: ok

chenc041: The action of  `ENTER` does not need to change,
And has add a test case for this situation

wenqi73 (approved):
LGTM.

PR #7229, opened 2022-01-21
Add the TemplateRef parameter to the constructor

etfired1:
The fix directive NzSpaceItemDirective complies with the Angular structured directive standard


simplejason:
@hsuanxyz Do we need to declare it?

Issue #7263, opened 2022-02-16
[Tabs] On mobile scroll is prevented on top of the tab title


nikolatasikj:
### Reproduction link
[https://codesandbox.io/s/elated-snow-75x55u?file=/src/app/app.module.ts&resolutionWidth=320&resolutionHeight=675](https://codesandbox.io/s/elated-snow-75x55u?file=/src/app/app.module.ts&resolutionWidth=320&resolutionHeight=675)

### Steps to reproduce
1. Open Code Sandbox link
2. Stay on the first tab (it have a long context)
3. Toggle device toolbar on your browser, so you can enable mobile scrolling feature
4. Focus your mouse on some of the title of the tab (Tab1, Tab2, or Tab3), click and drag to try to scroll down

### What is expected?
Page to scroll down

### What is actually happening?
Page is not scrolled down

| Environment | Info |
|---|---|
| ng-zorro-antd | 13.0.1 |
| Browser | Google Chrome (Version 98.0.4758.80) |

---
You can test the same thing with some examples from the tabs documentation link https://ng.ant.design/components/tabs/en. The important part is to use the mobile scroll feature (click and drag) which is enabled with the device toolbar console feature on google chrome, not scrolling with the mouse.

PR #7268, opened 2022-02-17
fix(module:tabs): mouse drag (scroll) on top of tabs title

nikolatasikj:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: [7263](https://github.com/NG-ZORRO/ng-zorro-antd/issues/7263)

## What is the new behavior?
If tabs are not scrollable, the browser will take a native scroll on mobile

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/tabs/public-api.ts
+++ b/components/tabs/public-api.ts
@@ -11,6 +11,7 @@ export { NzTabBodyComponent as ɵNzTabBodyComponent } from './tab-body.component
 export { NzTabNavOperationComponent as ɵNzTabNavOperationComponent } from './tab-nav-operation.component';
 export { NzTabScrollListDirective as ɵNzTabScrollListDirective } from './tab-scroll-list.directive';
 export { NzTabCloseButtonComponent as ɵNzTabCloseButtonComponent } from './tab-close-button.component';
+export { NzTabHoverDirective as ɵNzTabHoverDirective } from './tab-hover.directive';
 export * from './tab.component';
 export * from './tab.directive';
 export * from './tab-link.directive';


simplejason:
image

Please check above, maybe we can make it work well after dragend event as except :) After that, we can merge this PR.

nikolatasikj:
> image
> 
> Please check above, maybe we can make it work well after dragend event as except :) After that, we can merge this PR.

This is because of weird :hover behavior on drag events. I tried to escape that by implementing a custom hover directive that is with `mouseover` and `mouseout` events. At this event, I'm adding a custom `ant-tab-hover` class that has CSS for the hover effect. This class is with a fixed name because if the user changes the tab prefix.

components/tabs/style/index.less
```diff
@@ -165,7 +165,7 @@
       }
     }
 
-    &:hover {
+    &.ant-tab-hover {
```
simplejason, line 168: Please do not change any styles except `patch.less`, because these files are synced from `ant-design`.We can put it to `patch.less`.
Of course, if this style does not affect the final experience, we can also ignore it for now, after that we can approve and merge this PR :)

simplejason (changes requested):

components/tabs/style/index.less
nikolatasikj: We can add the new class in `patch.less`, but if we leave this as it is now (generated by `ant-design`) with `&:hover` it will still remain broken because this CSS on hover will trigger the old behavior and newly created directive will do nothing.

nikolatasikj: Should I add this class in `patch.less` and open PR in the `ant-design` repo or leave it as it is?

simplejason: You need to add these changes to patch.less like other components, after that we can merge this PR :)

nikolatasikj: Problem is that in `index.less` we have 
```
&:hover {
      color: @tabs-hover-color;
    }
```
so it will always override the added class from `patch.less`

Issue #7240, opened 2022-02-02
IconNotFoundError when using dynamically loaded icons with Angular Universal


nicholasgerard:
### Reproduction link
[https://github.com/NG-ZORRO/ng-zorro-antd/blob/f40aad9393930e7ff595cdf4679beb374414a572/components/icon/icon.directive.ts#L139](https://github.com/NG-ZORRO/ng-zorro-antd/blob/f40aad9393930e7ff595cdf4679beb374414a572/components/icon/icon.directive.ts#L139)

### Steps to reproduce
1. Create a project with Angular Universal
2. Attempt to use dynamically loaded SVG icons as described in docs with `nz-icon`
3. Build for SSR and run locally
4. App crashes

### What is expected?
Icons are loaded

### What is actually happening?
```
Angular is running on http://localhost:4000
Angular is running in development mode. Call enableProdMode() to enable production mode.
/Users/xxx/Documents/yyy/dist/zzz/server/main.js:309464
              throw error;
              ^

Error: [@ant-design/icons-angular]:the icon delete-o does not exist or is not registered.
    at IconNotFoundError (/Users/xxx/Documents/yyy/dist/zzz/server/main.js:477160:10)
    at /Users/xxx/Documents/yyy/dist/zzz/server/main.js:477315:15
```

| Environment | Info |
|---|---|
| ng-zorro-antd | 13.0.1 |
| Browser | Any |

---
See refactor of `changeIcon2`  in `NzIconDirective` 12.1.1:

https://github.com/NG-ZORRO/ng-zorro-antd/blob/f40aad9393930e7ff595cdf4679beb374414a572/components/icon/icon.directive.ts#L139

It appears this throws the above error when running in a Node environment (e.g. Angular Universal).

arturovt:
Hey @nicholasgerard . I've just tried the older version of the `nz-icon` in the simple skeleton app and I got the same error, because it can't load the SVG on the server-side (the `xhr2.XMLHttpRequest` throws an error that protocol is unsupported).

nicholasgerard:
Makes sense, thanks!

nicholasgerard:
This still seems to be an issue in 13.1.0.

arturovt:
[...]

PR #7290, opened 2022-03-02
fix(module:icon): do not try to load SVG on the Node.js side since it will throw an error

arturovt:
Issue Number: #7240


simplejason (approved):
LGTM

PR #7293, opened 2022-03-04
perf(module:modal): call `focus()` on the next rendering frame to prevent frame drop

arturovt:
Calling `focus()` on the element causes browsers to do re-layouts. This can been seen here:
https://gist.github.com/paulirish/5d52fb081b3570c81e3a#setting-focus
Microtasks are executed asynchronously, but within the current rendering frame. Using `requestAnimationFrame`
except of `Promise.resolve` basically unloads the current frame and calls `focus()` on the next rendering
frame, this prevent frame drops on slower devices when the modal is opened:

### Flame graph

image

### Reason

image


simplejason (approved):
LGTM

PR #7307, opened 2022-03-09
perf(module:tree-view): do not run change detection when the `nz-tree-node-checkbox` is clicked

arturovt:


simplejason (approved):
LGTM

simplejason (approved):
LGTM

PR #7313, opened 2022-03-10
perf(module:input-number): do not run change detection when `mouseup` and `mouseleave` events are dispatched on handlers

arturovt:


simplejason (approved):
LGTM

PR #7314, opened 2022-03-10
fix(module:list): re-enter the Angular zone when the `NgZone.onStable` emits

arturovt:
The `NgZone.onStable` always emits outside of the Angular zone, but the zone has not been
re-entered. This leads to change detection being called outside of the Angular zone.


cipchk:
LGTM

cipchk (approved):
LGTM

PR #7347, opened 2022-04-04
feat(module:avatar): add `loading` and `fetchpriority` attributes

### LINKED ISSUE #7274: Add loading="lazy" to Avatar img element
## What problem does this feature solve?
Defers loading the image until it reaches a calculated distance from the viewport, as defined by the browser. The intent is to avoid the network and storage bandwidth needed to handle the image until it's reasonably certain that it will be needed

## What does the proposed API look like?
	<nz-avatar [nzLoading]="'lazy'" [nzSrc]="src"></nz-avatar>

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

arturovt:
Closes #7274


simplejason:
As [caniuse->loading](https://caniuse.com/?search=loading) and [caniuse->fetchpriority](https://caniuse.com/?search=fetchpriority) shown, these attributes have low compatibility for most browsers, I don't think we need built-in properties to support. What do you think about this @hsuanxyz ?

arturovt:
@simplejason it's been a while since 2022, it's supported in all evergreen.

Laffery (approved):
LGTM

Issue #7656, opened 2022-09-30
nzScroll creates a vertical scrollbar in header


abika:
### Reproduction link
[https://angular-hjpeur.stackblitz.io](https://angular-hjpeur.stackblitz.io)

### Steps to reproduce
See the "Fixed Header" example in the official components documentations: 
https://ng.ant.design/components/table/en#components-table-demo-fixed-header

### What is expected?
No scrollbar in the table header

### What is actually happening?
Only (?) in Firefox there is an unnecessary vertical scrollbar in the table header.

| Environment | Info |
|---|---|
| ng-zorro-antd | 14.0.0 |
| Browser | Firefox 104 |

---
There is no header scrollbar in Chrome.

ousc:
I can submit a PR to fix it.

abika:
Many thanks! What is the status of the fix? I don't understand why the MR is closed.

Issue #7657, opened 2022-09-30
Table header and content misalignment


razvanbh:
### Reproduction link
[https://stackblitz.com/edit/angular-f6ptlu](https://stackblitz.com/edit/angular-f6ptlu)

### Steps to reproduce
1. Create a nz-table with `[nzScroll]="{ x: 'auto', y: '300px' }"`
2. Add a table header with nzWidth set and a long header text

### What is expected?
If header is wider than the set nzWidth the columns in the table take the table header width instead of the nzWidth value.

### What is actually happening?
The columns in the table are taking the width from nzWidth, but the header is getting wider, and the table is not aligned anymore with the header.
![image](https://user-images.githubusercontent.com/10511363/193310663-f7342c5f-6e6e-4645-997f-9c173bf30d54.png)

| Environment | Info |
|---|---|
| ng-zorro-antd | 14.0.0 |
| Browser | Chrome |

simplejason:
超长 title 可以使用 nzBreakWord 折行，或者手动更改宽度或样式

PR #7661, opened 2022-10-03
fix(module:table): fix vertical scroll bar in table header

ousc:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The table component of ng-zorro used to display a vertical scroll bar in the header, and modified its scroll bar style under webkit to achieve a similar display effect to Ant Design React, but this implementation uses styles that are only valid under webkit, Not compatible with firefox.
Issue Number: #7656 

## What is the new behavior?

After this modification, when the table has vertical scrolling, the header will add a scroll bar cell with the same width as the scroll bar of the table body in the last column of the first row and the last column of the table header like Ant Design React. If the column is fixed on the right, then this column is also fixed at the same time, and the corresponding right style is set for the other fixed columns on the right.

```
[ ] Yes
[x] No
```

## Other information
no

### DIFF
--- a/components/table/src/cell/cell-fixed.directive.ts
+++ b/components/table/src/cell/cell-fixed.directive.ts
@@ -26,6 +26,10 @@ export class NzCellFixedDirective implements OnChanges {
   isFixedRight = false;
   isFixed = false;
 
+  isInsideTableThead(): boolean {
+    return this.elementRef.nativeElement.closest('thead')?.contains(this.elementRef.nativeElement);
+  }
+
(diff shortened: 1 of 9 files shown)


ousc:
I just found out that #7657 may also be caused by this error, and maybe it can be solved through this PR, but I can't access the reproduction link provided, so I can't be sure.

ousc:
This PR needs to wait until #7672 Proceed after passing

PR #7801, opened 2023-01-05
fix(module:space): nzSpaceItem margin in rtl

akg2011-code:
- Change 'margin-right' to 'margin-inline-end' and
'margin-bottom' to 'margin-block-end' for support any direction


simplejason:
I find `margin-block-end` works only on v87+ in [caniuse](https://caniuse.com/?search=margin-block-end), I'm not sure if this is the best solution.

akg2011-code:
> I find `margin-block-end` only works on v87+ in [caniuse](https://caniuse.com/?search=margin-block-end), I'm not sure if this is the best solution.

What is the lowest version it should be?

lppedd:
> What is the lowest version it should be?

I think having proper localization (which includes RTL design) is more important than supporting old browser anyways.

Laffery:
> > What is the lowest version it should be?
> 
> I think having proper localization (which includes RTL design) is more important than supporting old browsers anyways.

Nearly two years on, we can look at the latest figures on [caniuse](https://caniuse.com/?search=margin-inline-end), both `margin-inline-end` and `margin-block-end` are supported by over 96% of browsers.
Besides, antd had already used `margin-inline-end` in their components https://github.com/search?q=repo%3Aant-design%2Fant-design+margin-inline-end&type=code.

I think it's time for us to import them in ng-zorro, WDYT? @lppedd

lppedd:
@Laffery yup, I agree. Generally speaking I'm ok with supporting older browsers, but I always prioritize the code and how easily a feature can be implemented with new options.

Laffery (approved):
LGTM

PR #8038, opened 2023-07-29
fix(module:tree): nzCheckBoxChange never emitting

CondensedMilk7:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `checkedKeysChange` never emits a value (`emit` is never called in the component), so two-way binding is not possible.

Issue Number: N/A

## What is the new behavior?

Upon checking or unchecking any node, the `checkedKeysChange` now emits all the checked keys present in tree nodes.

```
[ ] Yes
[x] No
```

## Other information

The test cases in general do not account for `EventEmitter` emissions. I can try to add a test for it if it is necessary.

Also let me know if there's anything else that needs to be changed or added.

### DIFF
--- a/components/core/tree/nz-tree-base.service.ts
+++ b/components/core/tree/nz-tree-base.service.ts
@@ -56,6 +56,26 @@ export class NzTreeBaseService {
   }
 
   /**
+   * get checked node keys
+   */
+  getCheckedNodeKeys(): NzTreeNodeKey[] {
+    const keys: NzTreeNodeKey[] = [];
+    const checkedNodes = this.getCheckedNodeList();
+
+    const calc = (nodes: NzTreeNode[]): void => {
+      nodes.forEach(node => {
+        keys.push(node.key);
+        if (node.children.length < 1) return;
+        calc(node.children);
+      });
+    };
+
+    calc(checkedNodes);
+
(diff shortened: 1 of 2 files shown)


CondensedMilk7:
It is also possible to pass `event.keys` to the EventEmitter in question, but this one works differently. Namely, if a node has all of its children selected, it will only contain the key of that node, instead of all of its children.

My solution will emit *all* the checked keys. Let me know whichever better suits the design.

CondensedMilk7:
@simplejason @hsuanxyz could you check this?

This bug looks like an important oversight.

Axel-Latour:
Hello @CondensedMilk7 
This PR is causing some regression. It doesn't take into account the `nzCheckStrictly`, so it's always sending all the children, even if you only want the parent node.
[Linked issue](https://github.com/NG-ZORRO/ng-zorro-antd/issues/8601)

CondensedMilk7:
@Axel-Latour I think [my comment above](https://github.com/NG-ZORRO/ng-zorro-antd/pull/8038#issuecomment-1657195802) went under the radar as nobody responded to it. It might be related.

PR #8064, opened 2023-08-23
fix(module:upload): remove inline style to resolve CSP issue

arturovt:
This commit addresses the CSP issue by removing the inline `style` property from the `input` element within the `nz-upload` component. The use of inline style attributes necessitates the inclusion of `unsafe-inline`.

### DIFF
--- a/components/upload/upload-btn.component.html
+++ b/components/upload/upload-btn.component.html
@@ -1,3 +1,8 @@
+<!--
+  We explicitly bind `style.display` to avoid using an inline style
+  attribute property (which is not allowed when CSP `unsafe-inline`
+  is not specified).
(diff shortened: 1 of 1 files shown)


HyperLife1119:
We should not add the new style directly to the antd style file, as this will interfere with the component style synchronization. We can either add the style to `patch.less` or use the style binding `[style.display]="'none'"`.

arturovt:
> We should not add the new style directly to the antd style file, as this will interfere with the component style synchronization. We can either add the style to `patch.less` or use the style binding `[style.display]="'none'"`.

Updated.

Nicoss54 (approved):
LGTM

PR #8065, opened 2023-08-23
fix(module:drawer): remove inline style to resolve CSP issue

arturovt:


HyperLife1119:
We should remove `--scroll-bar` from the drawer component because I can't find any code that is using it.

arturovt:
> We should remove `--scroll-bar` from the drawer component because I can't find any code that is using it.

Updated.

Nicoss54 (changes requested):
Thanks for the work, is it possible to make pass the job failed please ?

After that LGTM

Nicoss54 (approved):
LGTM

PR #8157, opened 2023-11-12
feat(module:image): support lazy loading for image component

ParsaArvanehPA:
## PR Checklist

- [✔] Tests for the changes have been added (for bug fixes / features)
- [✔] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[✔] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently when you add `loading='lazy'` to an image, it doesn't work as intended. It loads the image like as the page is loaded.

Issue Number: N/A

## What is the new behavior?
Now if you add `nzLoading='lazy'`, the image is only loaded when image is in viewport.

```
[ ] Yes
[✔] No
```


HyperLife1119:
Should we consider using `` directly?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading

ParsaArvanehPA:
> Should we consider using `` directly? https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading
> 
> NgOptimizedImage also uses the same technique https://angular.io/api/common/NgOptimizedImage#instance-properties

Unfortunately, this method is ineffective. The reason is that in `image.directive.ts`, the image element is created using `document.createElement('img')`. Once the element is created, its src attribute is immediately assigned, triggering the browser to download it. I attempted to set its loading attribute before setting its src, but it did not enable lazy loading.

I implement a manual approach to verify when the newly created img element appears in the viewport. I utilize `IntersectionObserver`, which ensures a high performance.

ParsaArvanehPA:
> > Should we consider using `` directly? https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading
> > NgOptimizedImage also uses the same technique https://angular.io/api/common/NgOptimizedImage#instance-properties
> 
> Unfortunately, this method is ineffective. The reason is that in `image.directive.ts`, the image element is created using `document.createElement('img')`. Once the element is created, its src attribute is immediately assigned, triggering the browser to download it. I attempted to set its loading attribute before setting its src, but it did not enable lazy loading.
> 
> I implement a manual approach to verify when the newly created img element appears in the viewport. I utilize `IntersectionObserver`, which ensures a high performance.
When the img element enters the viewport, its `src` attribute is assigned, causing it to be downloaded by the browser.

ParsaArvanehPA:
> Thanks for your contribution :) @ParsaArvanehPA This year we applied for some free licenses from Jetbrains(valid until Nov 14, 2024, it can be activated in All Products Pack of Jetbrains), please contact me if you need it, I've sent you a notification to your email!

Hi Jason,

Thank you so much for your wonderful message and your generous offer. I am very happy and honored to be part of the ng-zorro-antd open source community. I enjoy working on this project and I am glad that my code has been useful and helpful.

I am very grateful for the license from JetBrains that you have applied for me. This is a very valuable and thoughtful gift, and I appreciate your kindness and generosity. I have registered an account on JetBrains with my email address, and I've sent you my account info via email.

I look forward to using JetBrains products and continuing to support our ng-zorro-antd community. Thank you again for your support!

components/image/image.directive.ts
```diff
@@ -45,6 +47,7 @@ export class NzImageDirective implements OnInit, OnChanges, OnDestroy {
 
   @Input() nzSrc = '';
   @Input() nzSrcset = '';
+  @Input() nzLoading: nzLoading = 'eager';
```
simplejason, line 50: type `nzLoading` can be removed.
```diff
-- @Input() nzLoading: nzLoading = 'eager';
++ @Input() nzLoading?: 'eager' | 'lazy';
```

components/image/image.directive.ts
```diff
@@ -168,4 +189,11 @@ export class NzImageDirective implements OnInit, OnChanges, OnDestroy {
         });
     }
   }
+
+  private backLoadCompleteStateHandler(): void {
+    this.status = 'normal';
+    this.getElement().nativeElement.loading = this.nzLoading;
```
simplejason, line 195: This attribute `loading` is not supported by chrome browsers lower than version 77, so it's better to not add it if `nzLoadingType` is not set. (https://caniuse.com/?search=eager)

simplejason (changes requested):
Just left some small nits :)

components/image/image.directive.ts
ParsaArvanehPA: done.

components/image/image.directive.ts
ParsaArvanehPA: I apologize for the delay.
This could be removed entirely, as setting the image element's lazy loading does not produce the desired outcome.

ParsaArvanehPA:
Greetings @simplejason,
I am wondering if you have had a chance to look at the changes I made to my pull request based on your comments. I have been waiting for your feedback and approval for some time now. Is there anything else I need to do to move this process forward? Please let me know as soon as possible. Thank you for your time and assistance.

Laffery (approved):
LGTM, WDYT @HyperLife1119

PR #8037, opened 2023-07-29
feat(module:affix): support standalone component

HyperLife1119:
## Note

为了生成 standalone 标签，我增加了一种后缀，用于将组件/指令/管道标记为 standalone 的，目前暂定为 `:standalone`。

image

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Application (the showcase website) / infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #8182, opened 2023-11-22
feat(module:alert): support standalone component

evgeniyefimov:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

Transfrom `Alert` component to standalone simillar to https://github.com/NG-ZORRO/ng-zorro-antd/pull/8037

### DIFF
--- a/components/alert/alert.component.ts
+++ b/components/alert/alert.component.ts
@@ -3,7 +3,7 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { BidiModule, Direction, Directionality } from '@angular/cdk/bidi';
+import { Direction, Directionality } from '@angular/cdk/bidi';
 import { NgIf } from '@angular/common';
 import {
   ChangeDetectionStrategy,
@@ -37,7 +37,7 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'alert';
   exportAs: 'nzAlert',
   animations: [slideAlertMotion],
   standalone: true,
-  imports: [BidiModule, NgIf, NzIconModule, NzOutletModule],
+  imports: [NgIf, NzIconModule, NzOutletModule],
   template: `
     <div
       *ngIf="!closed"
--- a/components/alert/alert.component.ts
+++ b/components/alert/alert.component.ts
@@ -3,7 +3,8 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { Direction, Directionality } from '@angular/cdk/bidi';
+import { BidiModule, Direction, Directionality } from '@angular/cdk/bidi';
+import { NgIf } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
@@ -24,8 +25,10 @@ import { takeUntil } from 'rxjs/operators';
 
 import { slideAlertMotion } from 'ng-zorro-antd/core/animation';
 import { NzConfigKey, NzConfigService, WithConfig } from 'ng-zorro-antd/core/config';
+import { NzOutletModule } from 'ng-zorro-antd/core/outlet';
 import { BooleanInput } from 'ng-zorro-antd/core/types';
 import { InputBoolean } from 'ng-zorro-antd/core/util';
+import { NzIconModule } from 'ng-zorro-antd/icon';
 
 const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'alert';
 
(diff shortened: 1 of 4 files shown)


HyperLife1119 (approved):
LGTM

OriginRing:
Should the use of standalone be revealed in ''When To Use''? @HyperLife1119 @evgeniyefimov

HyperLife1119:
Adding two import methods at the same time may cause confusion for users. In order to resolve this confusion, we may also need to use a more detailed explanation, which will be more lengthy. I think just prompting the user that the component supports standalone usage would be enough, WDYT?

OriginRing:
> Adding two import methods at the same time may cause confusion for users. In order to resolve this confusion, we may also need to use a more detailed explanation, which will be more lengthy. I think just prompting the user that the component supports standalone usage would be enough, WDYT?

截屏2023-11-28 16 17 33
I added it in the hashcode before, I think it's okay, and 17 will default to standalone, I think we should also use standalone as the main one

HyperLife1119:
Looks good! Maybe we don't need the "Import Module" header anymore.

evgeniyefimov:
I think two types of imports in the docs may really confuse users. It'll be difficult to explain that users need to import multiple components for complex components to let them work correctly instead of a single module. Single module just more user friendly. I think it's better to mention only modules in the docs. It's simillar to Angular Material. Material's components and directives are standalone, but only module import is mentioned in the docs.

I would say that `standalone` components and directives have next purposes in the lib:
- simplify internal import dependencies managment
- provide alternative imports for advanced users (simillar to `CommonModule` and standalone directives from `@angular/common`)

OriginRing:
> I think two types of imports in the docs may really confuse users. It'll be difficult to explain that users need to import multiple components for complex components to let them work correctly instead of a single module. Single module just more user friendly. I think it's better to mention only modules in the docs. It's simillar to Angular Material. Material's components and directives are standalone, but only module import is mentioned in the docs.
> 
> I would say that `standalone` components and directives have next purposes in the lib:
> 
> * simplify internal import dependencies managment
> * provide alternative imports for advanced users (simillar to `CommonModule` and standalone directives from `@angular/common`)

For complex components, it is not recommended to support standalone and should be imported as module. @evgeniyefimov

MunMunMiao:
I believe that since the official direction is towards standalone components. So we can pursue the official position, which reduces everyone's communication costs and Zorro's future path.

I believe that shifting towards a comprehensive standalone and abandoning the Module is the direction for Angular's future.

Why not focus on future development now that there is community assistance available?

HyperLife1119:
It's worth noting that while Standalone is taking the lead, NgModule will not be deprecated, and it's entirely possible to have both (NgModule and Standalone), leaving it up to the user to decide whether they want to use an NgModule or a Standalone Component.
Personally, when a feature has a lot of components/instructions, I prefer to use NgModule because it's more cohesive and allows me to import all the components I need at once.
Of course, this can be a bit subjective, so the choice should be left to the user, all we need to do is to remind them that our components support standalone mode :)

ParsaArvanehPA:
The current documentation of the components indicates that they can also be standalone at the API section of each component. However, since standalone approach will be the default approach from Angular version 17, I think it would be better to change the order of the information. 
I suggest showing how to import the components as standalone at the beginning of each document, and then stating that they also support module at the API section. 
This way, we can align the documentation with the future direction of Angular development.

OriginRing:
The current standalone transformation has already started, so is it okay if we support standalone for all components in 17.1.0 and change the documentation and demos to standalone mode in one of the subsequent releases? @HyperLife1119 @MunMunMiao @simplejason

MunMunMiao:
> The current standalone transformation has already started, so is it okay if we support standalone for all components in 17.1.0 and change the documentation and demos to standalone mode in one of the subsequent releases? @HyperLife1119 @MunMunMiao @simplejason

Yes. Convert all components to standalone, with standalone as the main focus while retaining modules, ensuring that the component library remains consistent with the official development roadmap.
Removing modules will also become easier in the future. So it's not necessary to keep some modules and some standalone.

components/alert/alert.component.ts
```diff
@@ -24,15 +25,19 @@ import { takeUntil } from 'rxjs/operators';
...
+import { NzIconModule } from 'ng-zorro-antd/icon';
 
 const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'alert';
 
 @Component({
   selector: 'nz-alert',
   exportAs: 'nzAlert',
   animations: [slideAlertMotion],
+  standalone: true,
+  imports: [BidiModule, NgIf, NzIconModule, NzOutletModule],
```
Nicoss54, line 40: suggestion: You can remove BidModule because the directive exported by this module is not used here

Nicoss54 (changes requested):

Nicoss54 (approved):

Nicoss54:
LGTM

Nicoss54:
LGTM

PR #8185, opened 2023-11-26
feat(module:anchor): support standalone component

evgeniyefimov:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

Transfromed Anchor module to standalone simillar to https://github.com/NG-ZORRO/ng-zorro-antd/pull/8037

Seems `BidiModule` is unused. Didn't find any usage of this module in the anchor's module. Removed.

Imported only required directives from `@angular/common` instead of whole `CommonModule`.

### DIFF
--- a/components/anchor/anchor-link.component.ts
+++ b/components/anchor/anchor-link.component.ts
@@ -3,7 +3,8 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { Platform } from '@angular/cdk/platform';
+import { Platform, PlatformModule } from '@angular/cdk/platform';
+import { NgIf } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
@@ -26,6 +27,8 @@ import { NzAnchorComponent } from './anchor.component';
   selector: 'nz-link',
   exportAs: 'nzLink',
   preserveWhitespaces: false,
+  standalone: true,
+  imports: [NgIf, PlatformModule],
   template: `
     <a
       #linkTitle
--- a/components/anchor/anchor-link.component.ts
+++ b/components/anchor/anchor-link.component.ts
@@ -3,7 +3,7 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { Platform, PlatformModule } from '@angular/cdk/platform';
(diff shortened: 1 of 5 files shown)


HyperLife1119 (dismissed):
I don't think anchor should be a standalone component, because it usually requires the combination of multiple components (nz-anchor and nz-link) to be used. Anchor cannot be used alone.

All functions that require the combination of ***multiple*** components to function properly should not be standalone components. Instead, these components should be bundled into ***one*** feature through NgModule.

Based on this point of view, it is not suitable to be converted into a standalone component.

WDYT?

HyperLife1119:
I don't think anchor should be a standalone component, because it usually requires the combination of multiple components (nz-anchor and nz-link) to be used. Anchor cannot be used alone. Based on this point of view, it is not suitable to be converted into a standalone component. WDYT?

evgeniyefimov:
> I don't think anchor should be a standalone component, because it usually requires the combination of multiple components (nz-anchor and nz-link) to be used. Anchor cannot be used alone.
> 
> All functions that require the combination of _**multiple**_ components to function properly should not be standalone components. Instead, these components should be bundled into _**one**_ feature through NgModule.
> 
> Based on this point of view, it is not suitable to be converted into a standalone component.
> 
> WDYT?

Module is still there, so nz-anchor and nz-link bundled together when module is imported. Also docs mention module import. 

From user perspective nothing will be changed.

From developers point of view it will be easier to understand what dependencies are required in components and import only the necessary dependencies, thereby reducing the size of the bundle.

HyperLife1119:
I have changed my mind and you are right.  We should not restrict whether users should import ng modules or standalone components, this should be left to the user's choice. 👍

HyperLife1119 (approved):
LGTM

components/anchor/anchor-link.component.ts
```diff
@@ -26,6 +27,8 @@ import { NzAnchorComponent } from './anchor.component';
   selector: 'nz-link',
   exportAs: 'nzLink',
   preserveWhitespaces: false,
+  standalone: true,
+  imports: [NgIf, PlatformModule],
```
HyperLife1119, line 31: I discovered that PlatformModule is actually an empty module, let's delete it safely :)

PR #8240, opened 2023-12-01
feat(module:modal): support standalone component

ParsaArvanehPA:
## PR Checklist

- [✔] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[✔] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Updated Modal component to work as a standalone component.

```
[ ] Yes
[✔] No
```


components/modal/modal-confirm-container.component.ts
```diff
@@ -104,7 +109,19 @@ import { ModalOptions } from './modal-types';
...
+  },
+  imports: [
+    NgClass,
+    NgStyle,
+    NzPipesModule,
+    NzIconModule,
+    NgIf,
+    NzModalCloseComponent,
+    NzOutletModule,
+    PortalModule,
```
Nicoss54, line 121: suggestion: maybe in this case it's not needed to import all the PortalModule just to add the cdkPortalOutlet directive which is standalone. What do you think about importing only what we need to optimize the final bundle ?

components/modal/modal-container.component.ts
```diff
@@ -72,7 +76,18 @@ import { ModalOptions } from './modal-types';
...
     '(click)': 'onContainerClick($event)'
-  }
+  },
+  imports: [
+    NgClass,
+    NgStyle,
+    NgIf,
+    NzModalCloseComponent,
+    NzModalTitleComponent,
+    PortalModule,
```
Nicoss54, line 86: same here :)

components/modal/modal-footer.component.ts
```diff
@@ -69,7 +72,9 @@ import { ModalButtonOptions, ModalOptions } from './modal-types';
   host: {
     class: 'ant-modal-footer'
   },
-  changeDetection: ChangeDetectionStrategy.Default
+  changeDetection: ChangeDetectionStrategy.Default,
+  imports: [NgIf, NzOutletModule, NgForOf, NzButtonModule],
```
Nicoss54, line 76: suggestion: you can import NgFor instead ngForOf

Nicoss54 (changes requested):

components/modal/modal-confirm-container.component.ts
ParsaArvanehPA: Yeah i'm on the same page with importing only the required modules, but apparently `cdkPortalOutlet` is not standalone:
image

Nicoss54: > Yeah i'm on the same page with importing only the required modules, but apparently `cdkPortalOutlet` is not standalone: image

Mmmh; if you take a look at this link you can see that the directive is mark as standalone: 

https://github.com/angular/components/blob/main/src/cdk/portal/portal-directives.ts

ParsaArvanehPA: Yeah you are right, but it was added to version 17.1.0-next.1.

https://github.com/angular/components/blame/0762d69b394343555488b5aaada9444dd1a5083a/src/cdk/portal/portal-directives.ts#L74

ParsaArvanehPA: we are currently using `@angular/cdk": "^17.0.0`

Nicoss54: > Yeah you are right, but it was added to version 17.1.0-next.1.
> 
> https://github.com/angular/components/blame/0762d69b394343555488b5aaada9444dd1a5083a/src/cdk/portal/portal-directives.ts#L74

Ohh yeah true miss it !! Thanks. 

We will update this after. 

So LGTM

thanks for your work !!

Nicoss54:
LGTM

components/modal/modal-footer.component.ts
ParsaArvanehPA: Yeah you are right. changes been made.
But I find it a bit confusing as why NgForOf is suggested to be imported instead of NgFor...

Nicoss54: > Yeah you are right. changes been made. But I find it a bit confusing as why NgForOf is suggested to be imported instead of NgFor...

Yeah that’s something with your ide, don’t know which one you used and also the version 

but If you take a look at this link 

https://github.com/angular/angular/blob/main/packages/common/src/directives/ng_for_of.ts

You can see that NgForOf class is exported as NgFor

PR #8298, opened 2023-12-14
fix(module:calendar): style radio button not apply

Nicoss54:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In calendar component, radio have a specific class todisplay the radio input as a button. This class is applied when the NzRadioButtonDirective is injected

Unfortunatly we just import NzRadioButton so the NzRadiButton directive injection is always null and the correct class is not apply

This is due to the standalone migration

At the end NzRadioButtonDirective must have the NzRadioButton as a host directive

Issue Number: N/A

## What is the new behavior?

Correct radio button class is apply

```
[ ] Yes
[X] No
```

### DIFF
--- a/components/calendar/calendar-header.component.ts
+++ b/components/calendar/calendar-header.component.ts
@@ -17,7 +17,7 @@ import { FormsModule } from '@angular/forms';
 
 import { CandyDate } from 'ng-zorro-antd/core/time';
 import { DateHelperService, NzI18nService as I18n } from 'ng-zorro-antd/i18n';
-import { NzRadioComponent, NzRadioGroupComponent } from 'ng-zorro-antd/radio';
+import { NzRadioModule } from 'ng-zorro-antd/radio';
 import { NzSelectModule, NzSelectSizeType } from 'ng-zorro-antd/select';
 
 @Component({
@@ -63,7 +63,7 @@ import { NzSelectModule, NzSelectSizeType } from 'ng-zorro-antd/select';
     class: 'ant-fullcalendar-header',
     '[style.display]': `'block'`
   },
-  imports: [NzSelectModule, NgForOf, NgIf, FormsModule, NzRadioGroupComponent, NzRadioComponent],
+  imports: [NzSelectModule, NgForOf, NgIf, FormsModule, NzRadioModule],
   standalone: true
 })
 export class NzCalendarHeaderComponent implements OnInit {


Nicoss54:
@HyperLife1119 this PR is important because it fix a regression on UI due to the standalone migration. We can't release a new version of Ng-Zorro without this PR

ParsaArvanehPA:
In the `What is the current behavior` section you have said:
> At the end NzRadioButtonDirective must have the NzRadioButton as a host directive

If I understood your solution correctly:
It's not possible to set `NzRadioComponent` as a hostDirective for `NzRadioButtonDirective` since it is not a directive but a component.

The alternative is converting `NzRadioButtonDirective` to a component, and calling `nz-radio` in it's `template`, this new component should have all the inputs and outputs of  `NzRadioComponent` and needs to pass them to it in it's template.
I find this syntax extremely unsatisfying, is there any other way in your opinion? @Nicoss54 @HyperLife1119 

And btw imho, if the underlying problem is fixed, this pr will no longer be required, so maybe just wait for the solution.

Nicoss54:
> In the `What is the current behavior` section you have said:
> 
> > At the end NzRadioButtonDirective must have the NzRadioButton as a host directive
> 
> If I understood your solution correctly: It's not possible to set `NzRadioComponent` as a hostDirective for `NzRadioButtonDirective` since it is not a directive but a component.
> 
> The alternative is converting `NzRadioButtonDirective` to a component, and calling `nz-radio` in it's `template`, this new component should have all the inputs and outputs of `NzRadioComponent` and needs to pass them to it in it's template. I find this syntax extremely unsatisfying, is there any other way in your opinion? @Nicoss54 @HyperLife1119
> 
> And btw imho, if the underlying problem is fixed, this pr may no longer be required, so maybe just wait for the solution...

That’s true, the propose solution can’t work. I read too fast and don’t see that radio is a component. Thanks for your review !! :)  

So the second solution is not good enough too because it will be quite complicated to maintain for the future. 

By the way this PR can’t be closed and should be merge because currently there is a a regression on the calendar component. The radio button style is not applied. 

Let merge this PR when it is approved then we will find a solution to refactor correctly the system of NzRadioComponent, because if we release a minor version of the library with all the standalone components we will introduce a regression for sure and this should’nt be happen. Moreover this fix is just importing the NzRadioModule like it was the case when all the library was not in standalone component. So it’s just do a “rollback”

Nicoss54:
> > In the `What is the current behavior` section you have said:
> > > At the end NzRadioButtonDirective must have the NzRadioButton as a host directive
> > 
> > 
> > If I understood your solution correctly: It's not possible to set `NzRadioComponent` as a hostDirective for `NzRadioButtonDirective` since it is not a directive but a component.
> > The alternative is converting `NzRadioButtonDirective` to a component, and calling `nz-radio` in it's `template`, this new component should have all the inputs and outputs of `NzRadioComponent` and needs to pass them to it in it's template. I find this syntax extremely unsatisfying, is there any other way in your opinion? @Nicoss54 @HyperLife1119
> > And btw imho, if the underlying problem is fixed, this pr may no longer be required, so maybe just wait for the solution...
> 
> That’s true, the propose solution can’t work. I read too fast and don’t see that radio is a component. Thanks for your review !! :)
> 
> So the second solution is not good enough too because it will be quite complicated to maintain for the future.
> 
> By the way this PR can’t be closed and should be merge because currently there is a a regression on the calendar component. The radio button style is not applied.
> 
> Let merge this PR when it is approved then we will find a solution to refactor correctly the system of NzRadioComponent, because if we release a minor version of the library with all the standalone components we will introduce a regression for sure and this should’nt be happen. Moreover this fix is just importing the NzRadioModule like it was the case when all the library was not in standalone component. So it’s just do a “rollback”

Maybe a solution can be to remove a NzRadioButton Directive or make it deprecated first then have a new input for the NZRadioComponent called nzType which can be typed as default | button and apply style based on this input. The problem with that is: it will have breaking change for the user and a schematics need to be write.

HyperLife1119:
Here is a possible solution: https://stackblitz.com/edit/stackblitz-starters-wb3uaf

It no longer requires an NzRadioButtonDirective, but uses an Input to determine which selector is currently used.

WDYT?

Nicoss54:
> Here is a possible solution: https://stackblitz.com/edit/stackblitz-starters-wb3uaf
> 
> It no longer requires an NzRadioButtonDirective, but uses an Input to determine which selector is currently used.
> 
> WDYT?

Yeah this can work without any breaking change for the user. I can implement it in other one. 

But at this end this PR is still useful because we will remove the NzRadioButton directive so the NzRadioModule will import only NzRadioComponent and NzRadioGroupComponent and expert them. 

The calendar component need both so it make more sense to import the NzRadioModule instead importing the two component separately 

what do you thing ?

HyperLife1119 (approved):
LGTM

PR #8299, opened 2023-12-15
refactor(module:radio): refactor way to manage radio button style

Nicoss54:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[X] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently button style of radio is applied if we have the direcive nz-radio-button on element. Based on this directive, radio component check if an instance of NzRadioBtton directive is present thanks to the DI.

Issue Number: N/A

## What is the new behavior?

NzRadioButtonDirective is now removed, the button style is applied based on anInput with the alias nz-radio-button. With the way there is no breaking changewith the older way because it's totally transparent for the user. 

```
[ ] Yes
[X] No
```


Nicoss54:
@HyperLife1119  this is the PR to manage in other way the button style applied on a radio :)

HyperLife1119:
Instead of removing NzRadioButtonDirective, we should mark it as deprecated and then remove it in the next major release as it still belongs to the public API.

Nicoss54:
> he public API.

maybe i don't understand the public api, but in the documentation we list the selector right ? the selector nz-radio-button still exist that's why for me the documentation here don't change. Maybe I totally wrong. What do you think ?

HyperLife1119:
Because it is exported in public-api.ts, it belongs to the public API and we do not rule out users referencing it.
I rethought it, okay it's almost an empty class, let's delete it :)

Nicoss54:
> Because it is exported in public-api.ts, it belongs to the public API and we do not rule out users referencing it. I rethought it, okay it's almost an empty class, let's delete it :)

I remove the directive from the public-api.ts file :), so this directive is not expose anymore

https://github.com/NG-ZORRO/ng-zorro-antd/pull/8299/files#diff-205026341fd9d8dc942b5849d0dc5e835e672a8dc0859a1972cf65194b096c6d

Nicoss54:
@HyperLife1119 i think we could merge this one too, the directive is not exposed anymore by the public api file and as you said its an empty class. We will break nothing and we don't need to write a migration schematics.

What do you think ?

HyperLife1119 (approved):
LGTM

Issue #7304, opened 2022-03-09
nzSelect has white/empty space in compact theme


geesen:
### Reproduction link
[https://stackblitz.com/edit/ng-zorro-antd-ivy-4tffxv](https://stackblitz.com/edit/ng-zorro-antd-ivy-4tffxv)

### Steps to reproduce
 - Go to: https://ng.ant.design/components/select/en#components-select-demo-automatic-tokenization
 - Switch to "Compact Theme"
 - scroll down

### What is expected?
No white/empty space.
Compare with react in compact theme: https://ant.design/components/select/?theme=compact#components-select-demo-automatic-tokenization

### What is actually happening?
In compact theme there is an empty space after all options. It seems that the space increases with more options.
This only occurs when you need to scroll. 

| Environment | Info |
|---|---|
| ng-zorro-antd | 13.1.1 |
| Browser | Chrome 98, Edge 99 |

---
Maybe, the size/height of the CDK's virtual scroll viewport is calculated based on the option-height from the normal theme and not from the compact theme?

cipchk:
@geesen You can only add the `[nzOptionHeightPx]="24"` property in `nz-select` component. Currently, there is no relevant code that can only detect it.

vhphuc:
how do i set nzOptionHeigtPx to nz-select of nz-table pagination

PR #8361, opened 2024-01-18
fix(module:config):  add itemsize config for tree select table (#8347)

### LINKED ISSUE #8347: 官方文档select组件,切换紧凑主题查看下拉框，最后一条留白太大
### Reproduction link
[https://github.com/NG-ZORRO/ng-zorro-antd](https://github.com/NG-ZORRO/ng-zorro-antd)

### Steps to reproduce
官方文档select组件,切换紧凑主题查看下拉框，最后一条留白太大
image

### What is expected?
下拉框样式正常

不正常

| Environment | Info |
|---|---|
| ng-zorro-antd | 17.0.1 |
| Browser | chrome |

EnochGao:
fix #8347

### DIFF
--- a/components/core/config/config.ts
+++ b/components/core/config/config.ts
@@ -93,6 +93,7 @@ export interface SelectConfig {
   nzBorderless?: boolean;
   nzSuffixIcon?: TemplateRef<NzSafeAny> | string | null;
   nzBackdrop?: boolean;
+  nzOptionHeightPx?: number;
 }
 
 export interface AffixConfig {
@@ -295,6 +296,7 @@ export interface TableConfig {
   nzShowSizeChanger?: boolean;
   nzSimple?: boolean;
   nzHideOnSinglePage?: boolean;
+  nzVirtualItemSize?: number;
 }
 
 export interface TabsConfig {
@@ -329,6 +331,7 @@ export interface TreeConfig {
   nzBlockNode?: boolean;
   nzShowIcon?: boolean;
   nzHideUnMatched?: boolean;
+  nzVirtualItemSize?: number;
 }
 
 export interface TreeSelectConfig {
--- a/scripts/site/_site/doc/app/app.component.ts
+++ b/scripts/site/_site/doc/app/app.component.ts
@@ -12,6 +12,7 @@ import { environment } from '../environments/environment';
 import { AppService } from './app.service';
 import { ROUTER_LIST } from './router';
 import { loadScript } from './utils/load-script';
+import { NzConfigService } from 'ng-zorro-antd/core/config';
 
 interface DocPageMeta {
   path: string;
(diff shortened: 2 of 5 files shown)


scripts/site/_site/doc/app/app.component.ts
```diff
@@ -85,9 +86,20 @@ export class AppComponent implements OnInit {
...
       return;
     }
+    if (theme === 'compact') {
+      this.nzConfigService.set('select', {
+        nzOptionHeightPx: 28
+      });
+    } else {
+      this.nzConfigService.set('select', {
+        nzOptionHeightPx: 32
+      });
```
ParsaArvanehPA, line 96: Hi @EnochGao 
I believe this file is for the document site; setting global configs here works on your localhost, because you are developing in the document site, but other developers are not.

In here you have access to the theme type, because the flow button that you can see at the right bottom corner of NG-ZORRO site sets it; in our components we cannot identify what theme developers are using.

ParsaArvanehPA: This issue has been previously discussed in #7304, which confirms the same problem. @cipchk's comment agrees with my earlier remark that theme type detection is not feasible in components.

EnochGao: > Hi @EnochGao I believe this file is for the document site; setting global configs here works on your localhost, because you are developing in the document site, but other developers are not.
> 
> In here you have access to the theme type, because the flow button that you can see at the right bottom corner of NG-ZORRO site sets it; in our components we cannot identify what theme developers are using.

hi,Here's just fixing the site.The source code made no modifications, but instead added configuration properties to enable users to make global unified changes in their projects.

EnochGao: > This issue has been previously discussed in #7304, which confirms the same problem. @cipchk's comment agrees with my earlier remark that theme type detection is not feasible in components.

Because some projects switch themes dynamically, it is impractical for each component to change on its own

components/core/config/config.ts
```diff
@@ -329,6 +331,7 @@ export interface TreeConfig {
   nzBlockNode?: boolean;
   nzShowIcon?: boolean;
   nzHideUnMatched?: boolean;
+  nzVirtualItemSize?: number;
```
ParsaArvanehPA, line 334: You have made `nzVirtualItemSize` global for tab's config, but never changed tab component.

scripts/site/_site/doc/app/app.component.ts
ParsaArvanehPA: Oh Ok then; just please update the documentations so others would know that `nzVirtualItemSize` and `nzOptionHeightPx` are global configs.

EnochGao: > Oh Ok then; just please update the documentations so others would know that and are global configs.`nzVirtualItemSize``nzOptionHeightPx`

ok，My mistake. I'll add it right away

ParsaArvanehPA: And I suggest you follow the issue template; otherwise it might get closed. 
The github checks not getting started are probably caused by this as well; i remember those issues not following the template would get closed automatically.

EnochGao: > And I suggest you follow the issue template; otherwise it might get closed. The github checks not getting started are probably caused by this as well; i remember those issues not following the template used to get closed automatically.

thanks

Nicoss54:
@EnochGao  thanks for your PR! Great works, it will take me some time to review all the change

But again thanks for your work

Laffery (commented):
Basically LGTM, could you please rebase from master branch?

Issue #8353, opened 2024-01-11
nzIndex drawer


egalley:
### Reproduction link
[https://stackblitz.com/edit/angular-rntc7j?file=src%2Fapp%2Fapp.component.ts](https://stackblitz.com/edit/angular-rntc7j?file=src%2Fapp%2Fapp.component.ts)

### Steps to reproduce
1. Show a modal
2. Add a button in the modal to show the drawer (nzZIndex = 1001)

### What is expected?
Drawer should be over the modal like it used to work.

### What is actually happening?
Drawer is under the modal and cannot be accessed

| Environment | Info |
|---|---|
| ng-zorro-antd | 17.1.0 |
| Browser | Chrome / Edge / Opera |

ParsaArvanehPA:
Hi @egalley 
Thank you for submitting this issue;
If it's alright by everyone, I can start working on this issue.

In the meanwhile, you can use this hack to get it working in your code:
```ts
::ng-deep {
.cdk-global-overlay-wrapper:has(.ant-drawer) {
    z-index: 1001;
  }
}
```

Although mind the `has` pseudo-class [support](https://caniuse.com/css-has) as it is not that great when it comes to firefox.

egalley:
Thank you Parsa for taking in account this issue and to find me a temporary solution.

ParsaArvanehPA:
Greetings @egalley 

This pr should fix this issue #8373

PR #8373, opened 2024-01-27
fix(module:cdk): zIndex is not used properly when creating overlay

ParsaArvanehPA:
## PR Checklist

- [] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[✔] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The modal, drawer, or any other components that are made with the Angular CDK have the `zIndex` applied to the inner element of the created overlay. This is not desirable for two reasons:

1.  The created overlay is a direct child of the body element
2.  Each overlay has its own wrapper with the class name 'cdk-global-overlay-wrapper'
Therefore, changing the zIndex of the inner element has no effect, as it is already isolated from other overlays. The expected behavior is that the `zIndex` should be applied to the overlay wrapper instead.

The image below illustrates this issue:
-  The red sections are the overlay wrappers

-  The green section is the inner element of the overlay wrapper, which has its `zIndex` changed

The scenario is as follows:
We have a modal component, which contains a button that opens a drawer component (the first red outline is the drawer wrapper, the second is the modal wrapper).
As stated in the issue below, the drawer component should appear above the modal component; but currently, the modal component is on top. The reason why the zIndex property does not work is that it is applied to the inner element of the overlay, rather than the overlay wrapper.

This issue occurs whenever the CDK overlay is used; currently, only modal and drawer components have been fixed. If this pull request is approved, I will update other components that use the overlay as well.

Issue Number: #8353 , #8336

## What is the new behavior?
Mentioned above.

```
[ ] Yes
[✔] No
```

## Other information
Currently it is not possible to set overlays' zIndex value when creating it ([issue](https://github.com/angular/components/issues/1432)).

image

### DIFF
--- a/components/core/overlay/overlay-z-index.ts
+++ b/components/core/overlay/overlay-z-index.ts
@@ -1,0 +1,12 @@
+/**
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
+ */
+
+import { OverlayRef } from '@angular/cdk/overlay';
+
+export function overlayZIndexSetter(overlayRef: OverlayRef, zIndex?: number): void {
+  if (!zIndex) return;
+
+  (overlayRef['_host'] as HTMLElement).style.zIndex = `${zIndex}`;
+}
--- a/components/core/overlay/public-api.ts
+++ b/components/core/overlay/public-api.ts
@@ -6,3 +6,4 @@
 export * from './nz-connected-overlay';
 export * from './nz-overlay.module';
 export * from './overlay-position';
+export * from './overlay-z-index';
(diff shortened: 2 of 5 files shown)


Nicoss54 (changes requested):
Please could you provide test for the code added ?

ParsaArvanehPA:
> Please could you provide test for the code added ?

Sure, I will get back to you with that as soon as I can.

ParsaArvanehPA:
> Please could you provide test for the code added ?

Sorry for the delay; tests for the changes have been added. 
Thank you for reviewing my PRs @Nicoss54

Nicoss54 (approved):
LGTM

PR #8439, opened 2024-03-14
fix(module:list): remove NgZone dependency

arturovt:
This commit removes the `NgZone` dependency from the list item cell. With zone.js becoming optional, `onStable` won't emit any value when `provideZonelessChangeDetection` is used in the application config. Instead, it now listens to a subject emission when the content projection has been initialized.


HyperLife1119 (approved):
LGTM

PR #8440, opened 2024-03-14
fix(module:typography): remove NgZone dependency

arturovt:
This commit eliminates the `NgZone` dependency from the text edit component. Since zone.js is becoming optional, `onStable` won't emit any value when `provideZonelessChangeDetection` is used in the application config. Instead, we now use the alternative approach provided by `afterNextRender`. Most of the code that was using `onStable` should be replaced with `afterNextRender`.


HyperLife1119 (approved):
LGTM

Issue #8510, opened 2024-04-19
Docs issue: sometimes changing components not work


Laffery:
### Reproduction link

https://ng.ant.design/experimental/pipes/zh

### Steps to reproduce

way 1:
1. enter the doc site
2. change components, repeat as more as possible

way2:
1. visit https://ng.ant.design/experimental/pipes/zh
2. reload the window

### What is expected?
Nothing wrong

### What is actually happening?
1. error in console
```txt
chunk-NOKK6CHT.js:7 ERROR TypeError: t.hasAttribute is not a function
    at u7 (chunk-NOKK6CHT.js:7:28447)
    at FC (chunk-NOKK6CHT.js:7:48131)
    at RC (chunk-NOKK6CHT.js:7:48094)
    at NC (chunk-NOKK6CHT.js:7:48071)
    at Mo.create (chunk-NOKK6CHT.js:7:70206)
    at bf.createComponent (chunk-NOKK6CHT.js:7:73895)
    at new e (chunk-NOKK6CHT.js:13:247232)
    at e.ɵfac [as factory] (chunk-NOKK6CHT.js:13:247459)
    at fr (chunk-NOKK6CHT.js:7:23904)
    at WC (chunk-NOKK6CHT.js:7:51361)
```

3. failed to navigate to new component page and something wrong in the page:

![image](https://github.com/NG-ZORRO/ng-zorro-antd/assets/49607541/d4903c49-f130-41e8-a747-5e0a96419dbf)

| Environment | Info |
|---|---|
| ng-zorro-antd | 17.4.0 |
| Browser | Chrome latest |

Laffery:
https://github.com/angular/angular/issues/50133

Issue #8406, opened 2024-02-15
<nz-slider> can not work with SSR


msx15615634157:
### Reproduction link
[https://stackblitz.com](https://stackblitz.com)

### Steps to reproduce
after perform SSR rendering，the <nz-slider> will throw Exception  "TypeError: e.hasAttribute is not a function" and the circle can not move 

### What is expected?
no expection

### What is actually happening?
after perform SSR rendering，the <nz-slider> will throw Exception  "TypeError: e.hasAttribute is not a function" and the circle can not move 

| Environment | Info |
|---|---|
| ng-zorro-antd | 17.1.0 |
| Browser | google |

Issue #8181, opened 2023-11-21
nzTooltipTitle Error in ngHydration


s-triar:
### Reproduction link
-

### Steps to reproduce
1. use angular version 17 with SSR enabled
2. create main component and  in main.component.ts
```
import { Component } from '@angular/core';
import { CommonModule } from '@angular/common';
import { NzLayoutModule } from 'ng-zorro-antd/layout';
import { NzButtonModule } from 'ng-zorro-antd/button';
import { RouterModule } from '@angular/router';
import { NzIconModule } from 'ng-zorro-antd/icon';
import { NzGridModule } from 'ng-zorro-antd/grid';
import { NzSpaceModule } from "ng-zorro-antd/space";
import { NzToolTipModule } from 'ng-zorro-antd/tooltip';
@Component({
  selector: 'app-main',
  standalone: true,
  imports: [
    CommonModule, 
    NzLayoutModule,
    NzButtonModule,
    RouterModule,
    NzIconModule,
    NzGridModule,
    NzSpaceModule,
    NzToolTipModule
  ],
  templateUrl: './main.component.html',
  styleUrl: './main.component.scss'
})
export class MainComponent {

}

```

3. in main component html.
```
<nz-layout class="layout-main">
  <nz-header class="header-main">
    <span class="header-title-page">Toko</span>
  </nz-header>
  <nz-content id="content-main">
    <router-outlet></router-outlet>
  </nz-content>
  <nz-footer class="footer-navigation-main">
    <div nz-row nzJustify="space-around" nzAlign="middle">
      <div nz-col>
        <a
          nz-button
          nzSize="large"
          nzShape="circle"

          aria-label="Menu Home"
          role="navigation"
          nz-tooltip [nzTooltipTitle]="homeTooltip" nzTooltipPlacement="top" 
        >
          <span nz-icon nzType="home"></span>
        </a>
        <ng-template #homeTooltip>
          <span>Home</span>
        </ng-template>
      </div>
      <div nz-col>
        <a
          nz-button
          nzSize="large"
          nzShape="circle"

          aria-label="Menu Profile"
          role="navigation"
          nz-tooltip nzTooltipTitle="profileTooltip" nzTooltipPlacement="top" 
        >
          <span nz-icon nzType="user"></span>
        </a>
        <ng-template #profileTooltip>
          <span>Profile</span>
        </ng-template>
      </div>
      <div nz-col>
        <a
          nz-button
          nzSize="large"
[...]

PR #8512, opened 2024-04-19
fix(module:tooltip,popover,popconfirm): fix hydration error

HyperLife1119:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #8510, #8181, #8406

https://angular.dev/errors/NG0500

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/tooltip/base.ts
+++ b/components/tooltip/base.ts
@@ -5,10 +5,10 @@
 
 import { Direction, Directionality } from '@angular/cdk/bidi';
 import { CdkConnectedOverlay, ConnectedOverlayPositionChange, ConnectionPositionPair } from '@angular/cdk/overlay';
+import { isPlatformBrowser } from '@angular/common';
 import {
   AfterViewInit,
   ChangeDetectorRef,
-  ComponentRef,
   Directive,
   ElementRef,
   EventEmitter,
@@ -16,11 +16,14 @@ import {
   OnDestroy,
   OnInit,
   Optional,
+  PLATFORM_ID,
   Renderer2,
   SimpleChanges,
   TemplateRef,
+  Type,
   ViewChild,
-  ViewContainerRef
+  ViewContainerRef,
+  inject
 } from '@angular/core';
 import { Subject, asapScheduler } from 'rxjs';
 import { delay, distinctUntilChanged, filter, takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


components/tooltip/base.ts
```diff
@@ -116,13 +112,18 @@ export abstract class NzTooltipBaseDirective implements OnChanges, OnDestroy, Af
...
+  protected hostView = inject(ViewContainerRef);
+  protected renderer = inject(Renderer2);
+  protected noAnimation = inject(NzNoAnimationDirective, { host: true, optional: true });
+  protected nzConfigService = inject(NzConfigService);
+
+  constructor(protected componentType: Type<NzTooltipBaseComponent>) {
+    afterNextRender(() => {
+      this.createComponent();
+      this.registerTriggers();
+    });
```
HyperLife1119, line 125: Key: afterNextRender will only be executed in the client, which ensures that components are only dynamically created in the browser.

components/tooltip/base.ts
```diff
@@ -171,7 +167,8 @@ export abstract class NzTooltipBaseDirective implements OnChanges, OnDestroy, Af
    * Create a dynamic tooltip component. This method can be override.
    */
   protected createComponent(): void {
-    const componentRef = this.componentRef;
+    const componentRef = this.hostView.createComponent(this.componentType);
```
HyperLife1119, line 170: Key: non-eager dynamic creation of components.

components/tooltip/base.ts
```diff
@@ -116,13 +113,24 @@ export abstract class NzTooltipBaseDirective implements OnChanges, OnDestroy, Af
...
+
+  ngOnInit(): void {
+    afterNextRender(
+      () => {
+        this.createComponent();
+        this.registerTriggers();
+      },
+      { injector: this.injector }
+    );
+  }
```
HyperLife1119, line 133: Key: afterNextRender will only be executed in the client, which ensures that components are only dynamically created in the browser.

components/tooltip/base.ts
```diff
@@ -116,13 +114,21 @@ export abstract class NzTooltipBaseDirective implements OnChanges, OnDestroy, Af
...
+  protected platformId = inject(PLATFORM_ID);
+
+  constructor(protected componentType: Type<NzTooltipBaseComponent>) {}
+
+  ngAfterViewInit(): void {
+    if (isPlatformBrowser(this.platformId)) {
+      this.createComponent();
+      this.registerTriggers();
+    }
+  }
```
HyperLife1119, line 131: The reason for the hydration failure is that the host view node was removed using the native API when the component was dynamically created, resulting in the node not being found when hydrating.

While the issue can be resolved by using `ViewContainerRef.detach` instead of removeChild, this also results in components within the overlay not being able to correctly measure element sizes.

By controlling this part of the logic to be executed only in the client, we can temporarily solve this problem.

In conclusion, we should not remove the host view of a component after it has been rendered. The best solution here would be to use the `Overlay.create` API to render the component directly to the overlay container so that there is no need to `removeChild`.

Nicoss54 (approved):
LGTM

Laffery (approved):
LGTM too :)

PR #8620, opened 2024-07-01
fix(module:dropdown): unable to set bool attr to false

kerwin612:
fix(module:dropdown): unable to set bool attr to false

The issue of not being able to update the value of a `boolean` attribute to **false** externally needs to be corrected.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/dropdown/dropdown.directive.ts
+++ b/components/dropdown/dropdown.directive.ts
@@ -205,10 +205,10 @@ export class NzDropDownDirective implements AfterViewInit, OnDestroy, OnChanges 
     if (nzTrigger) {
       this.nzTrigger$.next(this.nzTrigger);
     }
-    if (nzVisible) {
+    if (nzVisible !== undefined) {
       this.inputVisible$.next(this.nzVisible);
     }
-    if (nzDisabled) {
+    if (nzDisabled !== undefined) {
       const nativeElement = this.elementRef.nativeElement;
       if (this.nzDisabled) {
         this.renderer.setAttribute(nativeElement, 'disabled', '');


components/dropdown/dropdown.directive.ts
```diff
@@ -205,10 +205,10 @@ export class NzDropDownDirective implements AfterViewInit, OnDestroy, OnChanges
     if (nzTrigger) {
       this.nzTrigger$.next(this.nzTrigger);
     }
-    if (nzVisible) {
+    if (nzVisible !== undefined) {
       this.inputVisible$.next(this.nzVisible);
     }
-    if (nzDisabled) {
+    if (nzDisabled !== undefined) {
```
HyperLife1119, line 211: Here, nzVisible and nzDisabled are of type SimpleChange, not boolean. So I guess we don't need to use undefined to determine that, WDYT?

kerwin612: `nzVisible` and `nzDisabled` receive **boolean** values externally, but when fetching their values from the `changes` object, there can be three possible scenarios: **undefined** (when the external value has not been changed), **true** (change to true from false), or **false** (change to false from true). However, if the code logic is simply `if(nzVisible)`, it will only process the case where it is **true**. When the external source intends to change it to **false**, it will not take effect as expected.

HyperLife1119: I think you misunderstood that when nzDisabled is changed to false, changes.nzDisabled is of type SimpleChange, not boolean, because it's changes.nzDisabled, not this.nzDisabled.

PR #8877, opened 2024-11-12
refactor(module:core): expose `fromEventOutsideAngular` utility

arturovt:
In this commit, we introduce a helper function, `fromEventOutsideAngular`, as a replacement for
`ngZone.runOutsideAngular + fromEvent`. Using this helper function simplifies things, as there is
no longer a need to inject the `NgZone` service everywhere.

Note that there are no functional changes—only formatting changes, as indentation has been
shifted to the left.

### DIFF
--- a/components/resizable/resizable.spec.ts
+++ b/components/resizable/resizable.spec.ts
@@ -1,5 +1,5 @@
 import { ApplicationRef, Component, ElementRef, NgZone, ViewChild } from '@angular/core';
-import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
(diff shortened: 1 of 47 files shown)


Laffery (approved):
Wonderful Job! Please Rebase

Laffery (approved):

components/resizable/resizable.spec.ts
```diff
@@ -93,13 +93,15 @@ describe('resizable', () => {
       zone.simulateZoneExit();
       fixture.detectChanges();
       fixture.detectChanges();
-      tick(16);
-      fixture.detectChanges();
-      expect(testComponent.width).toBe(600);
-      expect(testComponent.height).toBe(200);
-    }));
+      afterNextFrameRender(() => {
```
Laffery, line 96: It seems that `tick` does not work for event stream outside of angular, thus using `setTimeout` instead

Issue #8940, opened 2024-12-17
Ant Design splitter component support


sholvabai:
## What problem does this feature solve?
allow user use splitter component with ng-zorro.

## What does the proposed API look like?
https://ant.design/components/splitter

PR #8987, opened 2025-01-21
feat(module:splitter): add splitter component

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes

Issue Number: #8940

## What is the new behavior?

Add splitter component, see https://ant.design/components/splitter

- [ ] Yes
- [x] No


components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,335 @@
...
+  providers: [NzDestroyService],
+  host: {
+    class: 'ant-splitter',
+    '[class.ant-splitter-horizontal]': 'nzLayout() === "horizontal"',
+    '[class.ant-splitter-vertical]': 'nzLayout() !== "horizontal"'
+  }
+})
+export class NzSplitterComponent {
+  /** ------------------- Props ------------------- */
+  nzLayout = input<NzSplitterLayout>('horizontal');
```
lppedd, line 100: Now that inputs are Signals, does it make to declare everything as `readonly` as you did for outputs?

Laffery: You're right, although `readonly` isn't required to inputs, we can still use it to make sure type safe during runtime!

lppedd: Indeed. I also like a consistent standard - so `readonly` by default (where applicable) is how I'd do it.

components/splitter/splitter-bar.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  ariaNow = input.required<number>();
+  ariaMin = input.required<number>();
+  ariaMax = input.required<number>();
+  active = input(false);
+  resizable = input(true);
+  vertical = input<boolean>();
+  lazy = input(false);
+  constrainedOffset = input<number>();
+
+  readonly previewTransform = computed(() => {
```
lppedd, line 51: Another opinion: I see that you used `protected` for `resizeStartEvent`, but not for properties (those signals) used in templates. This is another candiate for a project-wide standard: should all template-only props/funs be marked as `protected`, or should all be `public`?

Laffery: We do not expect user to use those internally-used-only props/funs, which are unstable and may be changed/removed in some refactoring PRs.

We only promise that public props/funs declared at the documentation website are compatible in a minor version. So I think they shall be `protected`.
But sometimes it's not convenient to access them in unit test cases.

TBH, there is no consistant standard in this project.
We cannot get the acknownledge of how many internal apis is being used by user, what we can do as a popular component library is to keep `public` apis in stable and avoid exposing too many unstale internal apis.

lppedd: > TBH, there is no consistant standard in this project.

Yeah that's what I saw by reading the code.

I would go the route of **not** bothering with specifying `protected`, and instead keep everything public because:
1. sometimes users need access to unstable / internal declarations when NG-ZORRO doesn't behave as expected
2. you could alternatively mark unstable / internal declarations with TSDoc's `@internal`

   ```ts
   /** @internal */
   readonly previewTransform = computed(() => {  ... })
   ```
   to signify to consumers this is not something they should normally use.  
   Additionaly, `@internal` plays nicely with [stripInternal](https://www.typescriptlang.org/tsconfig/#stripInternal) in case in the future you want to get rid of them in the outputted types.

Laffery:
@lppedd I'd made changes as your suggestions

Laffery:
This pr is nearly completed, all the use cases of antd are implemented.

There are Todos as follows till now:

- [ ] Restore the previous size when expending a collapsed panel
- [ ] Observer the resize events of the 
- [ ] RTL support
- [ ] Unit test cases

I will add them next week, and I think they do not have a impact on your review of the code, so could you please take a review on this PR if you have time. @Nicoss54 @HyperLife1119 @OriginRing @lppedd

components/splitter/splitter-bar.component.ts
lppedd: Btw, up to you. I was mostly trying to understand if there is a project/team standard, especially now that components are being refactored or created to support signals.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+  min: string | number | undefined;
+  max: string | number | undefined;
+  postPercentMinSize: number;
+  postPercentMaxSize: number;
+}
+
+@Component({
+  selector: 'nz-splitter',
+  exportAs: 'nzSplitter',
+  preserveWhitespaces: false,
```
lppedd, line 48: Isn't setting `preserveWhitespaces: false` redundant?

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+        [class.ant-splitter-mask-vertical]="nzLayout() !== 'horizontal'"
+      ></div>
+    }
+  `,
+  imports: [NgTemplateOutlet, NzSplitterBarComponent],
+  providers: [NzDestroyService],
+  host: {
+    class: 'ant-splitter',
+    '[class.ant-splitter-horizontal]': 'nzLayout() === "horizontal"',
+    '[class.ant-splitter-vertical]': 'nzLayout() !== "horizontal"'
```
lppedd, line 101: I would be explicit here and use `nzLayout() === "vertical"`.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+export class NzSplitterComponent {
+  /** ------------------- Props ------------------- */
+  readonly nzLayout = input<NzSplitterLayout>('horizontal');
+  readonly nzLazy = input(false, { transform: booleanAttribute });
+  readonly nzResizeStart = output<number[]>();
+  readonly nzResize = output<number[]>();
+  readonly nzResizeEnd = output<number[]>();
+
+  readonly destroy$ = inject(NzDestroyService);
+  readonly elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
```
lppedd, line 113: inject-ed services or elements are best kept `private` imo.  
I'd also keep a consistent structure:
1. private properties
2. inputs
3. outputs
4. component state (props/functions)

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+
+  /** ------------------- Panels ------------------- */
+  // Get all panels from content children
+  readonly panels = contentChildren(NzSplitterPanelComponent);
```
lppedd, line 117: `panels` can be private since it's used inside of the component itself only.  
The same can be done for `containerSize` and `innerSizes`.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+  );
+  readonly innerSizes = linkedSignal({
+    source: this.panelProps,
+    computation: source => source.map(panel => panel.defaultSize)
+  });
+  readonly sizes = computed(() => {
+    let emptyCount = 0;
+    const containerSize = this.containerSize();
+    const innerSizes = this.innerSizes();
+    const sizes = this.panelProps().map((panel, index) => {
```
lppedd, line 148: Can the lambda function be extracted with a good name? It would help in readability imo.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+        postPercentMaxSize
+      } as PanelSize;
+    });
+
+    const totalPercentage = sizes.reduce((acc, { percentage }) => acc + (percentage ?? 0), 0);
+
+    if (totalPercentage > 1 && !emptyCount) {
+      // If total percentage is larger than 1, we will scale it down.
+      const scale = 1 / totalPercentage;
+      sizes.forEach(size => {
```
lppedd, line 192: `forEach` looks weird when modifying mutable element's properties.  
I think an indexed `for` loop would be clearer, and you could also extract the code snippet into its own function.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+   * The offset of preview position (lazy mode) when dragging the splitter bar.
+   * Constrained by the min and max size of the target panel.
+   */
+  readonly constrainedOffset = signal<number>(0);
+  /**
+   * Handle the resize start event for the specified panel.
+   * @param index The index of the panel.
+   * @param startPos The start position of the resize event.
+   */
+  protected startResize(index: number, startPos: [x: number, y: number]): void {
```
lppedd, line 228: Unless one debugs this function (but still) it's very hard to understand what it does.  
Maybe a short paragraph explaining the theory behind it would be beneficial to other maintainers.

components/splitter/typings.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
+ */
+
+export type NzSplitterLayout = 'horizontal' | 'vertical';
+export type NzSplitterCollapsible = boolean | NzSplitterCollapseOption;
+
+export interface NzSplitterCollapseOption {
+  start?: boolean;
+  end?: boolean;
```
lppedd, line 11: `readonly` candidate.

components/splitter/doc/index.en-US.md
```diff
@@ -0,0 +1,41 @@
+---
+category: Components
+type: Layout
+title: Splitter
+tag: 19.2.0
+cover: https://mdn.alipayobjects.com/huamei_7uahnr/afts/img/A*f0SISaETY0wAAAAAAAAAAAAADrJ8AQ/original
+---
+
+Split panels to isolate.
```
lppedd, line 9:
```suggestion
Splits the area into multiple horizontal or vertical sides.
```

components/splitter/doc/index.en-US.md
```diff
@@ -0,0 +1,41 @@
...
+
+## When To Use
+
+Can be used to separate areas horizontally or vertically. When you need to freely drag and adjust the size of each area.
+When you need to specify the maximum and minimum width and height of an area.
```
lppedd, line 14:
```suggestion
The Splitter component can be used to separate areas horizontally or vertically.  
You can freely drag the splitter to adjust the size of each area, while also being able to specify its maximum and minimum width and height.
```

components/splitter/doc/index.en-US.md
```diff
@@ -0,0 +1,41 @@
...
+
+```ts
+import { NzSplitterModule } from 'ng-zorro-antd/splitter';
+```
+
+## API
+
+### nz-splitter
+
+| 参数                | 说明                              | 类型                           | 默认值            | 全局配置 |
```
lppedd, line 24: Column titles should be in English.

components/splitter/splitter-bar.component.ts
```diff
@@ -0,0 +1,87 @@
...
+    <div
+      class="ant-splitter-bar-dragger"
+      [class.ant-splitter-bar-dragger-disabled]="!resizable()"
+      [class.ant-splitter-bar-dragger-active]="active()"
+      (mousedown)="resizeStartEvent($event)"
+      (touchstart)="resizeStartEvent($event)"
+    ></div>
+
+    @if (collapsible()?.start) {
+      <div class="ant-splitter-bar-collapse-bar ant-splitter-bar-collapse-bar-start" (click)="collapse.emit('start')">
```
lppedd, line 35: You already have a component function `resizeStartEvent` that emits `offsetStart` events.  
I would extract `collapse.emit(...)` into a component function too, to streamline where to look when browsing code.

lppedd (commented):
Left some minor comments / nitpicks.

Nicoss54:
@Laffery @HyperLife1119 does it have a guideline for refactoring the components to support signal ? It will be a great pleasure to help you on this huge reworking. 

Take a look and give my review tomorrow if it’s good for you ?

HyperLife1119:
@Nicoss54 If there is no special situation, existing components will not be migrated to signals.

Laffery:
> @Nicoss54 If there is no special situation, existing components will not be migrated to signals.

Agree. That would be a huge work and I don't think the payoff would be great, and there could be unexpected problems introduced. 
I think we can leave it as is for now, unless there are bottlenecks in feature implementation or performance.
For new components, we 100% recommend using signals.

Laffery:
> I think we can leave it as is for now, unless there are bottlenecks in feature implementation or performance. For new components, we 100% recommend using signals.

By the way, material still remains in the original implementation as well

components/splitter/splitter.component.ts
Laffery: you're right, `preserveWhitespaces` defaults to be `false`

components/splitter/splitter.component.ts
Laffery: I prefer the following order

1. inputs
2. outputs
3. private properties
4. component state (props/functions)

Puting inputs/outputs at the very top makes it clear to the person viewing the code how the component is used, just like a instruction manual

lppedd:
Converting to signals during a component's fix or feature addition might be useful to expose unexpected property writes from inside or outside the component itself. But apart from that I agree and I don't see the urgency (also considering signals are not strictly necessary for zoneless).

components/splitter/splitter.component.ts
lppedd: Consistency over order, so that looks good to me.

components/splitter/splitter-bar.component.ts
```diff
@@ -0,0 +1,92 @@
...
+        [style.transform]="preview ? previewTransform() : null"
+      ></div>
+    }
+
+    <div
+      class="ant-splitter-bar-dragger"
+      [class.ant-splitter-bar-dragger-disabled]="!resizable()"
+      [class.ant-splitter-bar-dragger-active]="active()"
+      (mousedown)="resizeStartEvent($event)"
+      (touchstart)="resizeStartEvent($event)"
```
HyperLife1119, line 32: Maybe we can use [Pointer Event](https://developer.mozilla.org/zh-CN/docs/Web/API/Pointer_events), which supports both mouse and touch.

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,480 @@
...
+  collapsible: Required<NzSplitterCollapseOption>;
+}
+
+@Component({
+  selector: 'nz-splitter',
+  exportAs: 'nzSplitter',
+  template: `
+    <ng-template>
+      <ng-content></ng-content>
+    </ng-template>
```
HyperLife1119, line 64: If not used, this should be removed :)

HyperLife1119 (changes requested):

components/splitter/splitter-bar.component.ts
Laffery: I tried it and the drag doesn't work very well with a touch screen. 
Currently antd uses `mousedown` and `touchstart` as well, I think we can keep aligning with them.

Laffery:
Tests added and all the TODOs done, waiting for kindly code review :)

components/splitter/splitter.component.ts
```diff
@@ -0,0 +1,563 @@
...
+
+  /** ------------------- Sizes ------------------- */
+  /**
+   * Observe the size of the container.
+   */
+  protected readonly containerBox = toSignal(
+    this.resizeObserver.observe(this.elementRef).pipe(
+      map(([item]) => item.target as HTMLElement),
+      map(el => ({ width: el.clientWidth, height: el.clientHeight })),
+      takeUntil(this.destroy$)
```
HyperLife1119, line 160: Remove `takeUntil(this.destroy$)`, toSignal will do it :)

HyperLife1119 (changes requested):

HyperLife1119 (approved):
LGTM

WwwHhhYran (approved):
LGTM

PR #8998, opened 2025-02-02
refactor: replace with `fromEventOutsideAngular` in other places

arturovt:


HyperLife1119:
Please fix the test :)

components/tabs/tab-scroll-list.directive.ts
```diff
@@ -164,12 +164,14 @@ export class NzTabScrollListDirective implements OnInit {
   };
 
   onOffset(x: number, y: number, event: Event): void {
-    this.ngZone.run(() => {
-      this.offsetChange.emit({
-        x,
-        y,
-        event
+    if (this.offsetChange.observed) {
```
HyperLife1119: This 'observed' property does not yet exist in [rxjs6](https://v6.rxjs.dev/api/index/class/Subject). But angular/core still supports rxjs6, we can use observers.length instead.
```suggestion
    if (this.offsetChange.observers.length) {
```

Laffery (approved):
LGTM

PR #9041, opened 2025-03-10
feat(module:pageheader): disable back button if no history

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently back button in Page Header component is never disabled even if there is no history entry

Issue Number: N/A

## What is the new behavior?

Back button is disabled is there is no entry in the history navigation

- [ ] Yes
- [X] No

### DIFF
--- a/components/page-header/page-header.component.ts
+++ b/components/page-header/page-header.component.ts
@@ -13,21 +13,23 @@ import {
   ContentChild,
   ElementRef,
   EventEmitter,
+  inject,
   Input,
   OnDestroy,
   OnInit,
   Output,
   TemplateRef,
-  ViewEncapsulation,
-  inject
+  ViewEncapsulation
 } from '@angular/core';
 import { Subject } from 'rxjs';
 import { map, takeUntil } from 'rxjs/operators';
 
+import { NzButtonModule } from 'ng-zorro-antd/button';
 import { NzResizeObserver } from 'ng-zorro-antd/cdk/resize-observer';
 import { NzConfigKey, NzConfigService, WithConfig } from 'ng-zorro-antd/core/config';
 import { PREFIX } from 'ng-zorro-antd/core/logger';
 import { NzOutletModule } from 'ng-zorro-antd/core/outlet';
+import { NzSafeAny } from 'ng-zorro-antd/core/types';
 import { NzIconModule } from 'ng-zorro-antd/icon';
 
 import { NzPageHeaderBreadcrumbDirective, NzPageHeaderFooterDirective } from './page-header-cells';
@@ -45,11 +47,18 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
         <!--back-->
         @if (nzBackIcon !== null) {
           <div (click)="onBack()" class="ant-page-header-back">
-            <div role="button" tabindex="0" class="ant-page-header-back-button">
+            <button
+              nz-button
+              role="button"
+              tabindex="0"
+              class="ant-page-header-back-button"
+              nzType="text"
+              [disabled]="!enableBackButton"
+            >
               <ng-container *nzStringTemplateOutlet="nzBackIcon; let backIcon">
                 <nz-icon [nzType]="backIcon || getBackIcon()" nzTheme="outline" />
               </ng-container>
-            </div>
+            </button>
           </div>
         }
 
@@ -91,9 +100,10 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
     '[class.ant-page-header-compact]': 'compact',
     '[class.ant-page-header-rtl]': `dir === 'rtl'`
   },
-  imports: [NzOutletModule, NzIconModule]
+  imports: [NzOutletModule, NzIconModule, NzButtonModule]
 })
 export class NzPageHeaderComponent implements AfterViewInit, OnDestroy, OnInit {
+  private location = inject(Location, { optional: true });
   readonly _nzModuleName: NzConfigKey = NZ_CONFIG_MODULE_NAME;
 
   @Input() nzBackIcon: string | TemplateRef<void> | null = null;
@@ -110,8 +120,7 @@ export class NzPageHeaderComponent implements AfterViewInit, OnDestroy, OnInit {
   compact = false;
   destroy$ = new Subject<void>();
   dir: Direction = 'ltr';
-
-  private location = inject(Location, { optional: true });
+  enableBackButton = true;
 
   constructor(
     public nzConfigService: NzConfigService,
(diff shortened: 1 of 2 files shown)


components/page-header/page-header.component.ts
```diff
@@ -45,11 +47,18 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
         <!--back-->
         @if (nzBackIcon !== null) {
           <div (click)="onBack()" class="ant-page-header-back">
-            <div role="button" tabindex="0" class="ant-page-header-back-button">
+            <button
+              nz-button
```
HyperLife1119, line 51: As far as I know, in antd4, the back button is a div instead of a button.

https://4x-ant-design.antgroup.com/components/page-header-cn/

Nicoss54: That's true, but i tought we could have a button because, we have to explicity set the role of the div as button. Button is more appropriate in this case no ? 

Anyway if you think it's better to have let's have a div, and i will rework a little bit this PR to have the correct style

HyperLife1119: While I also think there should be a button here, since we need to synchronize with antd, it can only be a div for now :)

Nicoss54: That's okey let me rework so :)

Nicoss54: @HyperLife1119 done :)

components/page-header/page-header.component.ts
```diff
@@ -45,7 +47,11 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
         <!--back-->
         @if (nzBackIcon !== null) {
           <div (click)="onBack()" class="ant-page-header-back">
-            <div role="button" tabindex="0" class="ant-page-header-back-button">
+            <div
+              role="button"
+              tabindex="0"
+              [class]="!enableBackButton ? 'ant-page-header-back-button__disabled' : 'ant-page-header-back-button'"
```
HyperLife1119, line 53: I'd like to hide the button when it's not available, instead of showing the disable style, since it doesn't exist in antd, WDYT?

components/page-header/page-header.component.ts
```diff
@@ -91,9 +97,10 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
     '[class.ant-page-header-compact]': 'compact',
     '[class.ant-page-header-rtl]': `dir === 'rtl'`
   },
-  imports: [NzOutletModule, NzIconModule]
+  imports: [NzOutletModule, NzIconModule, NzButtonModule]
```
HyperLife1119, line 100: Remove the NzButtonModule.

components/page-header/page-header.component.ts
HyperLife1119, line 53: I think we don't need to add disabled style, because it is not part of antd :)

components/page-header/page-header.component.ts
```diff
@@ -110,8 +117,7 @@ export class NzPageHeaderComponent implements AfterViewInit, OnDestroy, OnInit {
   compact = false;
   destroy$ = new Subject<void>();
   dir: Direction = 'ltr';
-
-  private location = inject(Location, { optional: true });
+  enableBackButton = true;
```
HyperLife1119, line 120:
```suggestion
  private enableBackButton = true;
```

components/page-header/page-header.component.ts
```diff
@@ -91,9 +97,10 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
     '[class.ant-page-header-compact]': 'compact',
     '[class.ant-page-header-rtl]': `dir === 'rtl'`
   },
-  imports: [NzOutletModule, NzIconModule]
+  imports: [NzOutletModule, NzIconModule, NzButtonModule]
 })
 export class NzPageHeaderComponent implements AfterViewInit, OnDestroy, OnInit {
+  private location = inject(Location, { optional: true });
```
HyperLife1119, line 103:
```suggestion
  private location = inject(Location);
```

components/page-header/page-header.component.ts
```diff
@@ -144,6 +157,9 @@ export class NzPageHeaderComponent implements AfterViewInit, OnDestroy, OnInit {
   }
 
   onBack(): void {
+    if (!this.enableBackButton) {
+      return;
+    }
```
HyperLife1119, line 162:
```suggestion
    if ((this.location.getState() as NzSafeAny)?.navigationId < 1) {
      return;
    }
```

components/page-header/page-header.component.ts
Laffery: +1

Nicoss54: let's do that, agree with this idea

components/page-header/page-header.component.ts
Nicoss54: If we do that we will have breaking change, maybe people use this component without using the routing and in this case, Location provider will not be found resulting to an injector error

components/page-header/page-header.component.ts
Nicoss54: why don't use the enabelBackButton ?

components/page-header/page-header.component.ts
HyperLife1119: As far as I know, Location comes from @angular/common and it is provided in the root.

Nicoss54: you are right: https://github.com/angular/angular/blob/main/packages/common/src/location/location.ts

let me implement your changes :)

package-lock.json
HyperLife1119, line 1: This seems like an unnecessary change, can you remove it?

Nicoss54: @HyperLife1119 removed

Nicoss54:
@HyperLife1119 have an issue with build on the pipeline but i can't reproduce in local, could you help or have an idea ?

components/page-header/page-header.component.ts
```diff
@@ -45,7 +45,7 @@ const NZ_CONFIG_MODULE_NAME: NzConfigKey = 'pageHeader';
         <!--back-->
         @if (nzBackIcon !== null) {
           <div (click)="onBack()" class="ant-page-header-back">
-            <div role="button" tabindex="0" class="ant-page-header-back-button">
+            <div [hidden]="!enableBackButton" role="button" tabindex="0" class="ant-page-header-back-button">
```
HyperLife1119, line 48: Do you think we need to hide the button when it is unavailable? @Laffery

Laffery: The `hidden` attribute hides the element but still occupies space, which causes a blank space to appear before the title and still be selected by pressing `tab`. It's weird. 
I suggest using '@if' instead

Nicoss54: Ohh yes true, i didn’t think that user can do tab and select the element. Let me rework

Laffery (approved):
LGTM

HyperLife1119 (approved):
LGTM

PR #9085, opened 2025-04-08
fix(module:input-number): fix parser when type not working

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently when type on the input number parser is not correctly applied

Issue Number: N/A

## What is the new behavior?

Parser is applied when user type something

- [ ] Yes
- [X] No

### DIFF
--- a/components/input-number/input-number.component.ts
+++ b/components/input-number/input-number.component.ts
@@ -9,6 +9,7 @@ import { DOWN_ARROW, ENTER, UP_ARROW } from '@angular/cdk/keycodes';
 import { NgTemplateOutlet } from '@angular/common';
 import {
   afterNextRender,
+  afterRenderEffect,
   booleanAttribute,
   ChangeDetectionStrategy,
   Component,
@@ -24,6 +25,7 @@ import {
   numberAttribute,
   OnInit,
   output,
+  Renderer2,
   signal,
   untracked,
   viewChild,
@@ -204,6 +206,7 @@ export class NzInputNumberComponent implements OnInit, ControlValueAccessor {
   private injector = inject(Injector);
   private focusMonitor = inject(FocusMonitor);
   private directionality = inject(Directionality);
+  private renderer = inject(Renderer2);
   private nzFormStatusService = inject(NzFormStatusService, { optional: true });
   private autoStepTimer: ReturnType<typeof setTimeout> | null = null;
   private defaultFormater = (value: number): string => {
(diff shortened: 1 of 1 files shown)


Nicoss54:
@HyperLife1119 i'm satisfied about this fix, found this fix dirty. 

After investigation it seems, that the property value of the input is not synchronise with the displayValue signal. I checked the property on the DOM to be sure.

I used the renderer Token in a afterRenderEffect callback to resynchronyse the displayValue with the property value of the input but i don't think it's a correct way to do this.

Maybe we can work together on this, i Initilize a draft for you to take a look

HyperLife1119:
https://github.com/user-attachments/assets/4c1868f5-729f-4210-a155-6b5354d8bbe1

I don't seem to find the situation where displayValue is out of sync with DOM. Can you describe the steps to reproduce it in detail?

Nicoss54:
> Kapture.2025-04-12.at.11.22.21.mp4 
> 
> I don't seem to find the situation where displayValue is out of sync with DOM. Can you describe the steps to reproduce it in detail?

Yes :) quite right, the investigation begins because the behavior is no longer the same as on the number legacy tax. 

When a parser is applied, the parser must be applied to each keyboard input. 

Let's imagine that the parser only accepts numbers. If you type 3a, the result will be 3. 

This is the case for the Legay, but not for the new one. 

The problem lies with the setInputWhenType function. 

Let's take the case of value 3. 

Initially, the displayValue signal is set to 3. 

The user types the letter a. The value passed to the function is therefore 3a.

The parser, which accepts only numbers, is applied. So we set the displayValue signal to 3 

Here the displayValue value does not change. The value of the displayValue signal is not set at each keyboard input. This means that the value of the input is 3a, but the value of the displayValue signal is 3 
no change in detection will therefore be raised when the parser is applied, so we end up with the value 3a in the input. 

A quick way to change this would be to set the displayValue signal value every time the user types something 

 You'll see that once the setInputWhenType method is called, the value of the displayValue signal will be 3, but the value of the input will be 3a.

HyperLife1119:
QQ_1744442428636

In fact, the behavior of the new input-number component is closer to antd.

Nicoss54:
> QQ_1744442428636
> 
> In fact, the behavior of the new input-number component is closer to antd.
> 
> If antd is used as the implementation standard, the parser behavior of the input-number legacy component is wrong.

Ohhh okey sorry for that and for the extra work. Will investigate more in the future to be sure ! Thanks a lot

Nicoss54:
Behavior was same as andt

PR #9133, opened 2025-05-28
refactor(module:modal): replace `const enum`

arturovt:
Angular has a local compilation mode, intended for use with the TypeScript compiler option `isolatedModules`, which disallows importing `const enum`. In such cases, we replace it with a `const` object.


HyperLife1119 (approved):
LGTM

PR #9145, opened 2025-06-04
refactor(module:anchor): use inject instead constructor

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/components/anchor/anchor-link.component.ts
+++ b/components/anchor/anchor-link.component.ts
@@ -9,9 +9,10 @@ import {
   ChangeDetectionStrategy,
   Component,
   ContentChild,
+  DestroyRef,
   ElementRef,
+  inject,
   Input,
-  OnDestroy,
   OnInit,
   Renderer2,
   TemplateRef,
@@ -53,7 +54,13 @@ import { NzAnchorComponent } from './anchor.component';
     class: 'ant-anchor-link'
   }
 })
-export class NzAnchorLinkComponent implements OnInit, OnDestroy {
+export class NzAnchorLinkComponent implements OnInit {
+  public elementRef = inject(ElementRef<HTMLElement>);
+  private anchorComp = inject(NzAnchorComponent);
+  private platform = inject(Platform);
+  private renderer = inject(Renderer2);
+  private readonly destroyRef = inject(DestroyRef);
+
   @Input() nzHref = '#';
   @Input() nzTarget?: string;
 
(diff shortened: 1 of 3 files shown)


components/anchor/anchor-link.component.ts
```diff
@@ -74,12 +81,11 @@ export class NzAnchorLinkComponent implements OnInit, OnDestroy {
...
   @ViewChild('linkTitle') linkTitle!: ElementRef<HTMLAnchorElement>;
 
-  constructor(
-    public elementRef: ElementRef,
-    private anchorComp: NzAnchorComponent,
-    private platform: Platform,
-    private renderer: Renderer2
-  ) {}
+  constructor() {
+    this.destroyRef.onDestroy(() => {
```
lppedd, line 85: There are cases like this one, where `destroyRef` is used exclusively in the constructor, where I feel a constructor parameter is always preferred, instead of keeping a reference on each component instance. Or a local `inject` call.

components/anchor/anchor.component.ts
```diff
@@ -121,15 +129,14 @@ export class NzAnchorComponent implements OnDestroy, AfterViewInit, OnChanges {
...
-    public nzConfigService: NzConfigService,
-    private scrollSrv: NzScrollService,
-    private cdr: ChangeDetectorRef,
-    private platform: Platform,
-    private renderer: Renderer2
-  ) {}
+  constructor() {
+    this.destroyRef.onDestroy(() => {
+      clearTimeout(this.handleScrollTimeoutID);
+      this.destroy$.next(true);
```
Laffery, line 136: why not use `takeUntilDestroyed` in `registerScrollEvent`?

Nicoss54: the method  registerScrollEvent is called twice, one in the afterViewInit lifecycle  and one in the onChanges lifecycle

so each time time the method is called, we have to unsubscribe to the previous subscription and re subscribe to avoid to create multiple subscription.

I agree this needs to be improve, but i think it's better to do it in another PR WDYT ?

What i have in mind is that the nzContainer should not change during the lifecycle of the component. So we could move this method only in the AfterViewInit lifecycle WDYT ?

Nicoss54: @Laffery @HyperLife1119 if you are agree with this idea I will move all the code in the AfterViewInit Life cycle 

this could have a breaking change for user who really made nzContainer dynamic

Laffery: I agree with you, it's wired to change nzContainer during the lifecycle of anchor component. WDYT @HyperLife1119

HyperLife1119: I suggest keeping the status quo, nzContainer is indeed dynamic.

HyperLife1119 (approved):
LGTM

PR #9172, opened 2025-06-08
refactor(module:core): use inject pattern

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


components/core/outlet/string_template_outlet.directive.ts
Laffery, line 1: We can rename this file to `string-template-outlet.directive.ts` to maintain consistent file name style. WDYT

components/core/overlay/nz-connected-overlay.ts
```diff
@@ -25,16 +25,14 @@ type Dimensions = Omit<ClientRect, 'x' | 'y' | 'toJSON'>;
   providers: [NzDestroyService]
```
Laffery: `NzDestroyService` is no longer used in this directive, I think we can remove this provider.

components/core/services/breakpoint.ts
```diff
@@ -44,23 +45,17 @@ export const siderResponsiveMap: BreakpointMap = {
...
+  private resizeService = inject(NzResizeService);
+  private mediaMatcher = inject(MediaMatcher);
 
-  constructor(
-    private resizeService: NzResizeService,
-    private mediaMatcher: MediaMatcher
-  ) {
+  constructor() {
     this.resizeService
       .subscribe()
```
Laffery, line 54: Aha it's wired that this method is named as `subscribe` but returns an Observerable. It's confused to subscribe twice.

Laffery: I think we should rename this method later

components/core/services/resize.ts
```diff
@@ -27,17 +29,10 @@ export class NzResizeService implements OnDestroy {
...
 
-  constructor(
-    private ngZone: NgZone,
-    private rendererFactory2: RendererFactory2
-  ) {
-    this.renderer = this.rendererFactory2.createRenderer(null, null);
-  }
-
-  ngOnDestroy(): void {
-    // Caretaker note: the `handler` is an instance property (it's not defined on the class prototype).
```
Laffery, line 38: I think we should keep this comment

Laffery (changes requested):

components/core/outlet/string_template_outlet.directive.ts
Nicoss54: I agree :), let's do that

components/core/services/breakpoint.ts
Nicoss54: yes, this could be done in an other PR

components/core/services/resize.ts
Nicoss54: get it :)

PR #9222, opened 2025-06-14
refactor(module:icon): use inject pattern

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/components/icon/icon.directive.ts
+++ b/components/icon/icon.directive.ts
@@ -6,19 +6,19 @@
 import {
   AfterContentChecked,
   ChangeDetectorRef,
+  DestroyRef,
   Directive,
   Input,
   NgZone,
   OnChanges,
-  OnDestroy,
   Renderer2,
   SimpleChanges,
   booleanAttribute,
   inject,
   numberAttribute
 } from '@angular/core';
-import { Subject, from } from 'rxjs';
-import { takeUntil } from 'rxjs/operators';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { from } from 'rxjs';
 
 import { IconDirective, ThemeType } from '@ant-design/icons-angular';
 
@@ -33,7 +33,11 @@ import { NzIconPatchService, NzIconService } from './icon.service';
     class: 'anticon'
   }
 })
-export class NzIconDirective extends IconDirective implements OnChanges, AfterContentChecked, OnDestroy {
+export class NzIconDirective extends IconDirective implements OnChanges, AfterContentChecked {
+  private readonly ngZone = inject(NgZone);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+  public readonly renderer = inject(Renderer2);
+  private destroyRef = inject(DestroyRef);
   cacheClassName: string | null = null;
   @Input({ transform: booleanAttribute })
   set nzSpin(value: boolean) {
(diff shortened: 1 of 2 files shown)


components/icon/icon.directive.ts
```diff
@@ -68,21 +72,9 @@ export class NzIconDirective extends IconDirective implements OnChanges, AfterCo
...
 
-  private destroy$ = new Subject<void>();
-
-  constructor(
-    private readonly ngZone: NgZone,
-    private readonly changeDetectorRef: ChangeDetectorRef,
-    public readonly iconService: NzIconService,
-    public readonly renderer: Renderer2
-  ) {
+  constructor(public readonly iconService: NzIconService) {
```
HyperLife1119, line 75:
```ts
constructor() {
  super(inject(NzIconService));
  ...
}
```

Nicoss54: You’re right.

Nicoss54: @HyperLife1119 after investigatiob we can't do that cause the iconService is used in other method, so remove it form constructor will raise an error after

HyperLife1119: Then it should be migrated to inject pattern.

```ts
public readonly iconService = inject(NzIconService);
```

Nicoss54: iconService is needed for super(). Unfortunattely if you use the inject pattern, iconservice needs to be call with the prefixer this in super declaration. but super needs to be called before accessing this. If we want to migrate correctly  in injection pattern we need to revew a little bit the icon directive frome the package @antd-design/icons-angular

this is a specific use case were we can't migrate to inject pattern so easily

HyperLife1119: WDYT?

```ts
public readonly iconService = inject(NzIconService);

constructor() {
  super(inject(NzIconService));
  ...
}
```

Nicoss54: That could work, but from my point of view it’s looks a little bit weird cause we inject « twice » it’s look like we absolutely want to use the inject pattern but in this case it’s not really appropriate. 

Of course i can refactor to use the solution you proposed :) 

@Laffery WDYT ?

Laffery: I agree that it's weired to inject twice.
What about 

```ts
providers: [{
    provide: IconService,
    useClass: NzIconService
}]

// ...
(this._iconService as NzIconService).createIconfontIcon(`#${this.iconfont}`)
```

Nicoss54: I’m not sure this will solve our problem of calling super. Wdut to update the IcinDirevtive which is extended to make it a host directive ?

lppedd: Migration is still not mandatory, so if ctor injection works nicely for this usecase, why spending too much time figuring out an alternative that might not be as clean?

Laffery: @lppedd +1. ctor injection is not really bad here,  it is propably even the best implementation here.

Nicoss54: I agree too! Even if i will refactor the directive it will be weird and it’s too much development juste for using the inject function. 

If all of you are ok i think we could approved this one and merge it

HyperLife1119 (approved):
LGTM

Issue #9229, opened 2025-06-17
能否为气泡确认框 NzPopconfirmModule 增加一个 nzCancelDisabled 属性？


LuoYingxiong:
## What problem does this feature solve?
https://ng.ant.design/components/popconfirm/zh#components-popconfirm-demo-promise

我希望点击「确认」执行异步等待时，把「取消」按钮设为“禁用” 状态，以防止用户在等待过程中误点「取消」 ；异步执行失败时，再把取消」按钮设为“启用” 状态；

目前官方没有提供 nzCancelDisabled 属性，难以实现该效果，能否在未来的版本中增加nzCancelDisabled 属性？？

![Image](https://github.com/user-attachments/assets/a37afaf3-5373-400b-a524-1b2b992a9945)

## What does the proposed API look like?
[nzCancelDisabled]

Laffery:
场景是合理的，antd 支持通过 cancelButtonProps 设置取消按钮的属性，zorro 目前仅支持设置 nzCancelText，在使用上是不够灵活的。
我们考虑在 20.x 版本中支持该特性，如果感兴趣欢迎贡献 ：）

Nicoss54:
@Laffery i could take this one if you are agree

Laffery:
> [@Laffery](https://github.com/Laffery) i could take this one if you are agree

That's great 👍

Laffery:
From my perspective, I think we should not add a property like `nzCancelDisabled`, rather add `nzCancelProps` `nzOkProps` to align with antd, 
otherwise we propably need to add lots of properties  such as nzCancelType nzCancelLoading nzCancelDanger in the future

Nicoss54:
@Laffery that’s a good point, and what i would suggest but this needs also a refactoring. Cause what i saw in antd is that cancelProps et okProps are object type of ButtonProps. 

I suggest to do this refactor also in this issue. By combining signal, linkedSignal and computed, i think the refactoring will be smooth 

WDYT ? Are you ok with this design?

Laffery:
> WDYT ? Are you ok with this design? 

that sounds nice

HCenggel:
我也有同样的需求

Nicoss54:
> > [@Laffery](https://github.com/Laffery) i could take this one if you are agree
> 
> That's great 👍

Let’s do that I come back from my vacation, Friday. I will work on it  after my return

Laffery:
[...]

PR #9245, opened 2025-06-24
feat(module:popconfirm): implement disable for cancel button

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9229

- [ ] Yes
- [X] No


components/popconfirm/popconfirm.ts
```diff
@@ -65,18 +70,60 @@ export class NzPopconfirmDirective extends NzTooltipBaseDirective {
...
   @Input({ transform: booleanAttribute }) nzCondition: boolean = false;
   @Input({ transform: booleanAttribute }) nzPopconfirmShowArrow: boolean = true;
   @Input() @WithConfig() nzPopconfirmBackdrop?: boolean = false;
   @Input() @WithConfig() nzAutofocus: NzAutoFocusType = null;
 
+  /**
+   * @Deprected v21
+   * please use the nzOkButton object input to describe option of the ok button
+   */
+  nzOkText = input<string | null>(null);
```
HyperLife1119, line 80: Switching from @Input to inputSignal is a breaking change and we need to consider it carefully.

Nicoss54: Yep, but in this specific case, there will be no regression. Initially those inputs was just necessary in the template to set the “state” of buttons 

I migrate them to make my own recalculation cause temporary we will have two system of how to pass the config object of action buttons and finally it’s the recalculation that is called in the template

All the previous tests pass as expected without I changed noting on it. 

I think this could be ship with a version 20.x of ng zorro

components/popconfirm/popconfirm-option.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
+ */
+
+import { NzButtonType } from 'ng-zorro-antd/button';
+
+interface NzPopConfirmButton {
+  nzType: NzButtonType | 'danger';
```
Laffery:
```suggestion
  nzType: NzButtonType | null;
```

components/popconfirm/popconfirm.ts
```diff
@@ -65,18 +70,60 @@ export class NzPopconfirmDirective extends NzTooltipBaseDirective {
...
+  nzCancelText = input<string | null>(null);
+
+  nzOkButtonProps = input<null | NzPopConfirmButtonProps>(null);
+  nzCancelButtonProps = input<null | NzPopConfirmButtonProps>(null);
+
+  /**
+   * TODO
+   * The following variables needs to be removed in v21 when we will remove the deprecated variable
+   * it v21 it will be not necessary anymore to maintain two system
+   * The goal is to have only the nzButton and nzCancelButton
```
Laffery:
```suggestion
   * The goal is to have only the nzOkButton and nzCancelButton
```

components/popconfirm/popconfirm-option.ts
Nicoss54: NZButtonType has already null in the tuple, but not danger. And to have no regression with the old system we need to have danger cause it check to the template. 

Before the change it was just a string, with this way it’s more strict

We need the danger mode in the type cause initilly we check this value in the input nzDanger of the ok button.

We could remove that but at this moment we change the initially behavior

Laffery: In the old system, there are both `nzOkType` and `nzOkDanger` existing individually, the migration path for user could be 

```
nzOkType --> nzOkButtonProps.nzType
nzOkDanger --> nzOkButtonProps.nzDanger

# but not
nzOkType --> nzOkButtonProps.nzType
nzOkDanger --> nzOkButtonProps.nzType='danger'
```

I think we don't need to add 'danger' here because the new system can cover the old, and it will also make inconsistance with `nz-button` and increase maintain cost.

- If user uses `nzOkType='danger'` with the old system, we can keep it compatible.
- If user uses the new system, I think it should be align with properties of `nz-button`.
- If user uses both, `nzOkType='danger'` and `nzOkButtonProps={xxx}`, we can map it to nzOkButtonProps.nzDanger to keep compatible. (we can drop these baggage in v21)

WDYT?


```suggestion
  nzType: NzButtonType;
```

Nicoss54: Mmmhhh right ! Get it now ! Will make sense. That’s true I made my own recalculation in the computed so i could map correctly instead complicated the type. 

I will make the change 

thanks

components/popconfirm/popconfirm.ts
```diff
@@ -65,18 +70,60 @@ export class NzPopconfirmDirective extends NzTooltipBaseDirective {
...
-  @Input() nzCancelText?: string;
   @Input() nzBeforeConfirm?: () => Observable<boolean> | Promise<boolean> | boolean;
   @Input() nzIcon?: string | TemplateRef<void>;
   @Input({ transform: booleanAttribute }) nzCondition: boolean = false;
   @Input({ transform: booleanAttribute }) nzPopconfirmShowArrow: boolean = true;
   @Input() @WithConfig() nzPopconfirmBackdrop?: boolean = false;
   @Input() @WithConfig() nzAutofocus: NzAutoFocusType = null;
 
+  /**
+   * @Deprecated v21
```
HyperLife1119:
```suggestion
   * @deprecated v21
```

components/popconfirm/popconfirm.ts
```diff
@@ -177,30 +221,26 @@ export class NzPopconfirmDirective extends NzTooltipBaseDirective {
                     #cancelBtn
                     [nzSize]="'small'"
                     (click)="onCancel()"
+                    [disabled]="nzCancelButtonProps()?.nzDisabled"
                     [attr.cdkFocusInitial]="nzAutoFocus === 'cancel' || null"
```
WwwHhhYran, line 206: Why not also support to set `nzType` and `nzDanger` of cancel btn, just like ok btn? 🤔 Although we usually dont change these properties of cancel bth, I think we can leave the choice to users. 

WDYT？

Nicoss54:
@HyperLife1119 @Laffery does anyone have an idea why some test not affected by last PRs  (https://github.com/NG-ZORRO/ng-zorro-antd/pull/9250) suddenly failed in the pipeline ? I take a look on the last pipeline ran on master and see there is the same problem. What is strange is that the pipeline not failed

components/popconfirm/popconfirm.ts
Nicoss54: I agree with that but it have to be done in an other PR, this PR was to disabel cancel button. But i can take the point in another PR

WDYT ?

components/popconfirm/popconfirm.ts
```diff
@@ -65,18 +70,60 @@ export class NzPopconfirmDirective extends NzTooltipBaseDirective {
...
-  @Input() nzCancelText?: string;
   @Input() nzBeforeConfirm?: () => Observable<boolean> | Promise<boolean> | boolean;
   @Input() nzIcon?: string | TemplateRef<void>;
   @Input({ transform: booleanAttribute }) nzCondition: boolean = false;
   @Input({ transform: booleanAttribute }) nzPopconfirmShowArrow: boolean = true;
   @Input() @WithConfig() nzPopconfirmBackdrop?: boolean = false;
   @Input() @WithConfig() nzAutofocus: NzAutoFocusType = null;
 
+  /**
+   * @deprecated v21
```
Laffery: Compare with antd

image

To align with antd, I think `nzOkText` `nzOkType` `nzCancelText` should be kept, and `nzOkDisabled` `nzOkDanger` should be deprecated and be replaced with `nzOkButtonProps` `nzCancelButtonProps`.
Therefore, `nzText` of `NzPopConfirmButton` should be removed cuz it's not the own property of `nz-button` but the content.

WDYT?

Nicoss54: Correct, let me rework the PR in this way

Nicoss54: @Laffery it's done and align with the antd design

HyperLife1119 (approved):
LGTM

PR #9301, opened 2025-07-28
refactor(module:float-button): migrate to signal

WwwHhhYran:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

## Other information

重构并统一 float button 为 signal 写法，以便于后期维护以及新增 tooltip, badge 功能。

Refactor  float button with signal for easier maintenance.

### DIFF
--- a/components/float-button/float-button-content.component.ts
+++ b/components/float-button/float-button-content.component.ts
@@ -42,7 +42,7 @@ import { NzIconModule } from 'ng-zorro-antd/icon';
   `
 })
 export class NzFloatButtonContentComponent {
-  nzIcon = input<string | TemplateRef<void> | null>(null);
-  nzDescription = input<string | TemplateRef<void> | null>(null);
-  nzShape = input<NzShapeSCType>('circle');
+  readonly nzIcon = input<string | TemplateRef<void> | null>(null);
+  readonly nzDescription = input<string | TemplateRef<void> | null>(null);
+  readonly nzShape = input<NzShapeSCType>('circle');
 }
--- a/components/float-button/float-button-top.component.ts
+++ b/components/float-button/float-button-top.component.ts
@@ -22,6 +22,7 @@ import {
   output,
   signal,
   TemplateRef,
+  untracked,
   viewChild,
   ViewEncapsulation
 } from '@angular/core';
@@ -93,9 +94,9 @@ export class NzFloatButtonTopComponent implements OnInit {
 
   // compact global config
   private readonly visibilityHeight = withConfig('nzVisibilityHeight', this.nzVisibilityHeight, 400);
-  public shape = linkedSignal(() => this.nzShape());
-  protected dir = this.directionality.valueSignal;
-  protected class = computed<string[]>(() => {
+  readonly shape = linkedSignal(() => this.nzShape());
+  protected readonly dir = this.directionality.valueSignal;
+  protected readonly class = computed<string[]>(() => {
     const dir = this.dir();
     const classes = [CLASS_NAME, `${CLASS_NAME}-top`, this.generateClass(this.shape())];
     if (dir === 'rtl') {
@@ -108,7 +109,7 @@ export class NzFloatButtonTopComponent implements OnInit {
   });
 
   private target?: HTMLElement | null = null;
-  private visible = signal<boolean>(false);
+  private readonly visible = signal<boolean>(false);
   private backTopClickSubscription = Subscription.EMPTY;
   private scrollListenerDestroy$ = new Subject<void>();
 
@@ -144,7 +145,7 @@ export class NzFloatButtonTopComponent implements OnInit {
 
     effect(() => {
       this.visibilityHeight();
-      this.handleScroll();
+      untracked(() => this.handleScroll());
     });
   }
 
--- a/components/float-button/float-button.component.ts
+++ b/components/float-button/float-button.component.ts
@@ -75,9 +75,9 @@ export class NzFloatButtonComponent {
   readonly nzShape = input<NzShapeSCType>('circle');
   readonly nzOnClick = output<boolean>();
 
-  public shape = linkedSignal(() => this.nzShape());
-  protected dir = inject(Directionality).valueSignal;
-  protected class = computed<string[]>(() => {
+  readonly shape = linkedSignal(() => this.nzShape());
+  protected readonly dir = inject(Directionality).valueSignal;
+  protected readonly class = computed<string[]>(() => {
     const dir = this.dir();
     const classes = [CLASS_NAME, this.generateClass(this.shape())];
     if (dir === 'rtl') {


components/float-button/float-button-top.component.ts
```diff
@@ -49,102 +53,107 @@ const passiveEventListenerOptions = normalizePassiveListenerOptions({ passive: t
...
 
-  readonly _nzModuleName: NzConfigKey = NZ_CONFIG_MODULE_NAME;
-
-  private scrollListenerDestroy$ = new Subject<void>();
-  private target?: HTMLElement | null = null;
+  readonly backTop = viewChild('backTop', { read: ElementRef });
 
-  visible: boolean = false;
-  dir: Direction = 'ltr';
+  @Input({ transform: numberAttribute }) @WithConfig() nzVisibilityHeight: number = 400;
```
Nicoss54: @Laffery i'm not sure about that, but is it not possible to inject the config and initialize an input signal with this injection. I know that the decorator @WithConfig will set the value property with the default value of the theme, but with signals using this way to do will be complicated

suggestion is to inject the config and set the signal value with the config or default value here 400 

WDYT

Nicoss54 (approved):
LGTM

Nicoss54:
@Laffery when you have time could you review ? :)

components/float-button/float-button-top.component.ts
Laffery: AGREE 


```suggestion
  private @WithConfig() _defaultVisibilityHeight = 400;
  nzVisibilityHeight = input(this._defaultVisibilityHeight, { transform: numberAttribute });
```

WwwHhhYran: 🤔 If the above method is adopted, the property in `BackUpConfig` will be declared as `_defaultVisibilityHeight`, which is inconsistent with the Config interface declarations of other components ( properties usually start with `nz`)。At the same time, this action may cause breaking changes？

Why can't we continue to maintain the current situation? Although different from `inputSignal`. 

Alternatively, we can inject `NZ_CONFIG` to set the default value for inputSignal ？In the future, we can even remove the `@WithConfig` decorator

@Laffery WDYT ?

Nicoss54: I’m in favor of injection NZ_CONFIG token and set the input signal with the default value of the theme :)

Laffery: In fact, `@WithConfig` is used not only when initializing the class field, but also when the input value changes. If we just **inject NZ_CONFIG to set the default value for inputSignal**, that would be a breaking change.

I think `@WithConfig` is similar to `transform` function of input, what about creating a new transfrom function? I'm thinking about it

Laffery: The question is: 

Once the `NZ_CONFIG` changes, `onConfigChangeEventForComponent` will mark for CD and then you will get the updated config value, because `@WithConfig` have hijacked the getter of the field.

But it's difficult to udpate the inputSignal, which is not writable.

Nicoss54: @Laffery sorry my late answer, busy days actually at work. After re reading the code in depth you're write

What we can do maybe is to rewritte the config to used signal and then combined input and linkedSignal. Input first is initialized with the config

and then linkedSignal recalculate the correct vew value in case input change or config change ?

WDYT ?

Laffery: You're right.
`transformFn` does not support this scenario well, `linkedSignal` is better I think

Laffery: A simple sample, I'm not sure whether it works or not, just for example. BTW I think the usage is not elegant because we should add an additional protect field

```ts
export function inputWithConfig<K extends NzConfigKey, T>(
  componentName: K,
  inputSignal: InputSignalWithTransform<T, unknown>
): WritableSignal<T> {
  ngDevMode && assertInInjectionContext(inputWithConfig);

  const nzConfigService = inject(NzConfigService);
  const configValue = nzConfigService.getConfigForComponent(componentName);

  return linkedSignal({
    source: inputSignal,
    computation: (value, previous) => {
      const assignedByUser = value !== undefined;

      if (assignedByUser) {
        return value;
      }

      if (configValue !== undefined) {
        return configValue as T;
      }

      return previous?.value as T;
    }
  });
}
```

usage:

```ts
nzVisibilityHeight = input(400, { transform: numberAttribute });
protect readonly _nzVisibilityHeight = inputWithConfig(NZ_CONFIG_MODULE_NAME, this.nzVisibilityHeight)
```

Nicoss54: true, we have more logic which need more maintainance and we duplicate variable finally. 

Second things if the config value change during runtime, linkedValue is not recalculated with this implementation right? Which could be a breaking change right ?

Laffery: Right, this is the issue of this implementation, which is not expected. 

I'm trying to design a implementation to avoid breaking changes. Welcome any advice!

Nicoss54: @Laffery maybe i could help ? we could have a sort a brain storming ? I have the idea to pass an tuple Signal which combine the input and the config signal ?

Laffery: yes. 

What I think:
merge the two signals -> computeSignal -> as source of the linkedSignal

WwwHhhYran: @Nicoss54 @Laffery I have created `withConfigSignal` to replace the `@WithConfig` decorator. Do you have any suggestions? :)

Nicoss54: This works :) amazing work. Maybe instead of start with void 0 we can start with the default value of the config ? This will avoid each time the config change to recalculate the value in the updateConfig function ? WDYT?

Laffery: @WwwHhhYran great job. I think the API could be much more simple. I implement it with decorators and you just need to bind the input field name.

```ts
type SignalValue<T> = T extends Signal<infer U> ? U : T;

export function WithConfigSignal<This, Value extends Signal<T>, T = SignalValue<Value>>(name: keyof This) {
  return function (_value: undefined, context: ClassFieldDecoratorContext<This, Value>) {
    context.addInitializer(function () {
      const nzConfigService = inject(NzConfigService);
      const inputSignal = this[name] as InputSignal<T>;
      const originalValue = inputSignal();
      const configKey = this['_nzModuleName' as keyof This] as NzConfigKey;
      const configValueSignal = toSignal<T | undefined>(
        nzConfigService
          .getConfigForComponentChange(configKey)
          .pipe(map(config => config?.[name as keyof NzConfig[NzConfigKey]]))
      );

      const value = computed<T>(() => {
        const configValue = configValueSignal();
        const inputValue = inputSignal();
        const assignedByUser = inputSignal[SIGNAL].version > 0 && isDefined(inputValue);

        if (assignedByUser) {
          return inputValue;
        } else if (configValue !== undefined) {
          return configValue;
        }

        return originalValue;
      });

      Object.defineProperty(this, context.name, {
        get: () => value,
        enumerable: true,
        configurable: false
      });
    });
  };
}
```

You can use it like this:

```ts
  nzSize = input<NzSizeLDSType>('default');

  @WithConfigSignal('nzSize') protected readonly size!: Signal<NzSizeLDSType>;
```

that's all.

I'm going to create a new PR about it and write enough test cases for it. @Nicoss54

Laffery: #9374

Laffery: @WwwHhhYran #9374 was merged, you can rebase from master and redo

WwwHhhYran:
@Nicoss54 @Laffery I have updated the code. Could you please code review again? :)

components/float-button/float-button-top.component.ts
```diff
@@ -49,102 +52,112 @@ const passiveEventListenerOptions = normalizePassiveListenerOptions({ passive: t
...
+    effect(() => {
+      const target = this.nzTarget();
+      if (target) {
+        this.target = typeof target === 'string' ? this.doc.querySelector(target) : target;
+        this.registerScrollEvent();
+      }
+    });
+
+    effect(() => {
+      const backTop = this.backTop();
```
Nicoss54, line 131: I think we could use the cleanup fonction like this


```typescript
effect((onCleanup) => {
  const backTop = this.backTop();
  if(backTop) {
      this.backTopClickSubscription = fromEventOutsideAngular(backTop.nativeElement, 'click')
          .pipe(takeUntilDestroyed(this.destroyRef))
          .subscribe(() => {
            this.scrollSrv.scrollTo(this.getTarget(), 0, { duration: this.nzDuration() });
            this.ngZone.run(() => this.nzOnClick.emit(true));
          });
       }

     return onCleanup(() => {
          this.backTopClickSubscription.unsubscribe() 
      })


})
```

WDYT ?

Nicoss54: it's more a suggestion than a change. But @WwwHhhYran great works!

components/core/config/config.service.ts
```diff
@@ -204,8 +203,8 @@ export function withConfigFactory<T extends NzConfigKey>(componentName: T) {
     return computed<V>(() => {
       const configValue = configValueSignal()?.[name] as V | undefined;
       const inputValue = inputSignal();
-      // if the version of the inputSignal is 0 or the inputValue is undefined, we consider it as not assigned by user
-      const assignedByUser = inputSignal[SIGNAL].version > 0 && isDefined(inputValue);
+      // if the inputValue is undefined, we consider it as not assigned by user
+      const assignedByUser = isDefined(inputValue);
```
WwwHhhYran: I found an issue：

If removing `inputSignal[SIGNAL]. version > 0`, `assignedByUser` will always be true. When user modifies value in config (use the `set` func of `NzConfigService`), the value will not be used and the computed signal still returns 400, which is not as expected. 

However, when using the `@WithConfig` decorator, if the user has not set `nzVisibilityHeight`, the config value will be prioritized, and the default value (400) will only be used if config value does not exist.

If we maintain `inputSignal[SIGNAL]. version > 0` condition, if the `nzVisibilityHeight` input by the user is exactly equal to defaultValue (400), the signal will not update when encountering the same value, so its version will still be 0. At this time, `assignedByUser` is false, which is not as expected too.

For the above issue, perhaps we can address it through the following method:
```ts
nzVisibilityHeight = input<number>();
// the originValue of withConfigFactory will be replace with the third arg
this.visibilityHeight = withConfig('nzVisibilityHeight', this.nzVisibilityHeight, 400)
```
This method can solve the above issue, but for community developers, they don not know should set the default value in `withConfig` instead of `inputSignal`. Do you have any ideas about this ?
@Nicoss54 @Laffery @HyperLife1119 @OriginRing

HyperLife1119: This issue does exist. If the user manually sets the input value to the default value, VERSION will still be equal to 0, causing the assignedByUser condition to be false, thus assigning the configuration value.

This can be solved by moving the default value to withConfig , setting the initial value of input() to undefined. Once the user actively sets the input value, VERSION will always be greater than 1 (unless the default value is also undefined 😅).

Regarding the community developer's question, I believe this can be resolved by modifying the withConfig function signature to only accept an InputSignal with undefined ( `InputSignal<V | undefined>`).

```ts
function isDefined<T>(value?: T): value is Exclude<T, undefined> {
  return value !== undefined;
}

export function withConfigFactory<T extends NzConfigKey>(componentName: T) {
  return <N extends keyof NonNullable<NzConfig[T]>, V = NonNullable<NzConfig[T]>[N]>(
    name: N,
    inputSignal: InputSignal<V | undefined> | InputSignalWithTransform<V | undefined, unknown>,
    defaultValue: V
  ): Signal<V> => {
    const configValueSignal = inject(NzConfigService)['_getConfigValue'](componentName);

    return computed<V>(() => {
      const configValue = configValueSignal()?.[name] as V | undefined;
      const inputValue = inputSignal();
      // if the version of the inputSignal is 0 or the inputValue is undefined, we consider it as not assigned by user
      const assignedByUser = inputSignal[SIGNAL].version > 0 && isDefined(inputValue);

      if (assignedByUser) {
        return inputValue;
      } else if (configValue !== undefined) {
        return configValue;
      }

      return defaultValue;
    });
  };
}

const withConfig = withConfigFactory('button');

@Directive()
export abstract class NzButton {
  nzSize = input<NzSizeLDSType>('default')
  size = withConfig('nzSize', this.nzSize, 'default');
                           // ^^^^^^^^^^^
                           // Error: Argument of type 'InputSignal<NzSizeLDSType>' is not assignable to parameter of type 'InputSignal<NzSizeLDSType | undefined> 
  nzSize = input<NzSizeLDSType>()
  size = withConfig('nzSize', this.nzSize, 'default'); // ok
}

```

Let me know what you think :) @Laffery

Laffery: Loooks good:)

Laffery: Let me do the change. cc @WwwHhhYran

components/float-button/float-button-top.component.ts
WwwHhhYran: good idea! I have implemented it :)

components/float-button/float-button-content.component.ts
```diff
@@ -41,7 +42,7 @@ import { NzIconModule } from 'ng-zorro-antd/icon';
   `
 })
 export class NzFloatButtonContentComponent {
-  @Input() nzIcon: string | TemplateRef<void> | null = null;
-  @Input() nzDescription: string | TemplateRef<void> | null = null;
-  @Input() nzShape: 'circle' | 'square' = 'circle';
+  nzIcon = input<string | TemplateRef<void> | null>(null);
+  nzDescription = input<string | TemplateRef<void> | null>(null);
+  nzShape = input<NzShapeSCType>('circle');
```
HyperLife1119:
```suggestion
  readonly nzIcon = input<string | TemplateRef<void> | null>(null);
  readonly nzDescription = input<string | TemplateRef<void> | null>(null);
  readonly nzShape = input<NzShapeSCType>('circle');
```

components/float-button/float-button-top.component.ts
```diff
@@ -49,102 +52,118 @@ const passiveEventListenerOptions = normalizePassiveListenerOptions({ passive: t
...
+  readonly nzTemplate = input<TemplateRef<void> | null>(null);
+  readonly nzTarget = input<string | HTMLElement | null>(null);
+  readonly nzDuration = input(450, { transform: numberAttribute });
+  readonly nzOnClick = output<boolean>();
+
+  // compact global config
+  private readonly visibilityHeight = withConfig('nzVisibilityHeight', this.nzVisibilityHeight, 400);
+  public shape = linkedSignal(() => this.nzShape());
+  protected dir = this.directionality.valueSignal;
+  protected class = computed<string[]>(() => {
```
HyperLife1119:
```suggestion
  readonly shape = linkedSignal(() => this.nzShape());
  protected readonly dir = this.directionality.valueSignal;
  protected readonly class = computed<string[]>(() => {
```

components/float-button/float-button-top.component.ts
```diff
@@ -49,102 +52,118 @@ const passiveEventListenerOptions = normalizePassiveListenerOptions({ passive: t
...
-  }
+    if (!this.visible()) {
+      classes.push(this.generateClass('hidden'));
+    }
+    return classes;
+  });
 
-  private doc = inject(DOCUMENT);
+  private target?: HTMLElement | null = null;
+  private visible = signal<boolean>(false);
```
HyperLife1119:
```suggestion
  private readonly visible = signal<boolean>(false);
```

components/float-button/float-button-top.component.ts
```diff
@@ -49,102 +52,118 @@ const passiveEventListenerOptions = normalizePassiveListenerOptions({ passive: t
...
+      }
+      return onCleanup(() => {
+        this.backTopClickSubscription.unsubscribe();
+      });
+    });
+
+    effect(() => {
+      this.visibilityHeight();
+      this.handleScroll();
+    });
```
HyperLife1119:
```suggestion
    effect(() => {
      this.visibilityHeight();
      untracked(() => this.handleScroll());
    });
```
If handleScroll does not need to be captured by the effect, consider placing it in untracked.

components/float-button/float-button.component.ts
```diff
@@ -3,90 +3,90 @@
...
+  readonly nzOnClick = output<boolean>();
 
-  ngOnInit(): void {
-    this.directionality.change?.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(direction => {
-      this.dir = direction;
-      this.cdr.detectChanges();
-    });
+  public shape = linkedSignal(() => this.nzShape());
+  protected dir = inject(Directionality).valueSignal;
+  protected class = computed<string[]>(() => {
```
HyperLife1119:
```suggestion
  readonly shape = linkedSignal(() => this.nzShape());
  protected readonly dir = inject(Directionality).valueSignal;
  protected readonly class = computed<string[]>(() => {
```

PR #9345, opened 2025-08-22
feat(module:modal): support nzLoading

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Modal support now nzLoading input and display a skeleton of loading when nzLoading is set to true

- [ ] Yes
- [X] No


components/modal/modal-container.component.ts
```diff
@@ -41,11 +42,17 @@ import { NzModalTitleComponent } from './modal-title.component';
...
-            <div [innerHTML]="config.nzContent"></div>
+          <div [hidden]="config.nzLoading">
+            <ng-template cdkPortalOutlet />
+            @if (isStringContent) {
+              <div [innerHTML]="config.nzContent"></div>
+            }
+          </div>
+          @if (config.nzLoading) {
+            <nz-skeleton [nzActive]="true" />
           }
```
HyperLife1119, line 53:
```suggestion
          @if (config.nzLoading) {
            <nz-skeleton [nzActive]="true" />
          } @else {
            <ng-template cdkPortalOutlet />
            @if (isStringContent) {
              <div [innerHTML]="config.nzContent"></div>
            }
          }
```

```diff
- @ViewChild(CdkPortalOutlet, { static: true })
+ @ViewChild(CdkPortalOutlet)
```

We also need to align antd in the DOM structure, so try not to add extra DOM nodes :)

Nicoss54: If we do that change at the opening of the modal it will raise an error which is normal cause the cdkPortalOutlet will not be present in the DOM structure, so it become impossible to attach the view
 
This change risk to impact much more things and make the code more complicated just to avoid an extra div

WDYT ?

Nicoss54: @HyperLife1119 WDYT about this solution

```html
        <div class="ant-modal-body" [style]="config.nzBodyStyle!" [hidden]="config.nzLoading">
          <ng-template cdkPortalOutlet />
          @if (isStringContent && !config.nzLoading) {
            <div [innerHTML]="config.nzContent"></div>
          }
        </div>
        <div class="ant-modal-body" [hidden]="!config.nzLoading">
          <nz-skeleton [nzActive]="true" />
        </div>
 ```
 
 that's true we will have two same div in the DOM with the same class, but this will avoid having an extra div as you mention inside the div having the class ant-modal-body. 
 

Using @if @else would result to 

- lazy create the content of the modal in function of loading or not. This will have a huge impact on the modal-content.directive file and how we will create the modal with service 

- an other solution was to keep always the CDKPortalOutlet directive and replace at runtime its content displayed 
But depending how huge is the content this could result to performance effect 

that’s my investigation until now :)

even if we duplicate the div, I don’t think this could have an impact in the future ? Even if antd design change. Or maybe I miss something in my analyses :)

Comment (Nicoss54) [reply components/modal/modal-container.component.ts:53]: @Laffery could we have also your point of view ? :)

Comment (Nicoss54) [reply components/modal/modal-container.component.ts:53]: @HyperLife1119 and @Laffery dis you take the time to analyses my idea ? :)

Comment (HyperLife1119) [reply components/modal/modal-container.component.ts:53]: This is still not a "clean" solution, and it seems the problem is a bit tricky. 🧐

Comment (Nicoss54) [reply components/modal/modal-container.component.ts:53]: Yeah this solution is more a workaround that a clean solution I agree :/ 

using if else too avoid  having an extra div result to make the problem more complex unfortunately.

Comment (Laffery) [reply components/modal/modal-container.component.ts:53]: @Nicoss54 I think the performance effect here is not too big. I prefer @HyperLife1119 's suggestion.
please refer to the antd's implementation https://github.com/ant-design/ant-design/pull/48848/files#diff-a3edcd43b644d0539a67f65e68ba30d6acb77a12531a6a71d1e26a008a9d752c

The key question of using `hidden` from my perspecitive is: when `nzContent` is a component, it will be initialized once the modal popup, even if the `nzLoading` is true.

In my opinion, the advantage of `nzLoading` is to add a transition period for user to query data asynchrously from the server. `nzContent` probably depends on the data from the server. 
So I think it's better to use `@if else` here.

Comment (Nicoss54) [reply components/modal/modal-container.component.ts:53]: Using @if @else is exactly the same approach than And design :) I need to review how to the modal content is create to have this feature.

Basically here just having @if block will not work due to the portal outlet directive and the way how we create the modal

Comment (HyperLife1119) [reply components/modal/modal-container.component.ts:53]: The real problem here is that using @if/else will make the type of portalOutlet nullable, which has a big impact and requires many changes. 

```diff
- @ViewChild(CdkPortalOutlet, { static: true }) portalOutlet: CdkPortalOutlet
+ @ViewChild(CdkPortalOutlet) portalOutlet: CdkPortalOutlet | null
```

Components like drawer also need to support nzLoading and may face the same problem, so a code refactoring is necessary.

Nicoss54: Exactly. For the moment I have to think how to refactor this and make you a proposition.

Nicoss54: @Laffery @HyperLife1119 i will close this PR, since a new one with te refactoring is necessary WDYT ?

HyperLife1119: Absolutely.

Nicoss54:
Need a big refactoring in the structure of modal and also potentially in drawer

PR #8826, opened 2024-10-08
feat(module:carousel): arrows for switching

ParsaArvanehPA:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Application (the showcase website) / infrastructure changes
- [ ] Other... Please describe:

## What is the new behavior?
Adding the ablility to have arrows for switching, like [Ant Design](https://ant.design/components/carousel#carousel-demo-arrows)

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #9355, opened 2025-08-26
feat(module:carousel): support nzArrows

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Carousel component now support nzArrows input which let the user to display arrow for swipping

- [ ] Yes
- [ ] No

## Other information

Related: #8826

### DIFF
--- a/components/carousel/carousel.spec.ts
+++ b/components/carousel/carousel.spec.ts
@@ -423,6 +423,80 @@ describe('carousel custom strategies', () => {
   }));
 });
 
+describe('carousel arrows', () => {
+  let fixture: ComponentFixture<NzTestCarouselArrowsComponent>;
+  let testComponent: NzTestCarouselArrowsComponent;
+  let carouselWrapper: DebugElement;
+  let carouselContents: DebugElement[];
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(NzTestCarouselArrowsComponent);
+    fixture.detectChanges();
+    testComponent = fixture.debugElement.componentInstance;
+    carouselWrapper = fixture.debugElement.query(By.directive(NzCarouselComponent));
+    carouselContents = fixture.debugElement.queryAll(By.directive(NzCarouselContentDirective));
+  });
+
+  it('should render arrows when nzArrows is true', () => {
+    const prev = carouselWrapper.nativeElement.querySelector('.slick-prev');
+    const next = carouselWrapper.nativeElement.querySelector('.slick-next');
+    expect(prev).not.toBeNull();
+    expect(next).not.toBeNull();
+  });
+
+  it('should navigate via arrows', fakeAsync(() => {
+    expect(carouselContents[0].nativeElement.classList).toContain('slick-active');
+    carouselWrapper.nativeElement.querySelector('.slick-next').click();
+    tickMilliseconds(fixture, 700);
+    expect(carouselContents[1].nativeElement.classList).toContain('slick-active');
+    carouselWrapper.nativeElement.querySelector('.slick-prev').click();
+    tickMilliseconds(fixture, 700);
+    expect(carouselContents[0].nativeElement.classList).toContain('slick-active');
+  }));
+
+  it('should disable arrows at edges when loop is false', fakeAsync(() => {
+    testComponent.loop = false;
+    fixture.detectChanges();
+    const prev = carouselWrapper.nativeElement.querySelector('.slick-prev');
+    const next = carouselWrapper.nativeElement.querySelector('.slick-next');
+    expect(prev.classList).toContain('slick-disabled');
+    expect(next.classList).not.toContain('slick-disabled');
+
+    // Go to last slide
+    for (let i = 0; i < 3; i++) {
+      next.click();
+      tickMilliseconds(fixture, 700);
+    }
+    expect(carouselContents[3].nativeElement.classList).toContain('slick-active');
+    expect(next.classList).toContain('slick-disabled');
+
+    // Clicking next should not move beyond last
+    next.click();
+    tickMilliseconds(fixture, 700);
+    expect(carouselContents[3].nativeElement.classList).toContain('slick-active');
+  }));
+});
+
+describe('carousel no swipe', () => {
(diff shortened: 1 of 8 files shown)


components/carousel/style/rtl.less
```diff
@@ -17,22 +17,24 @@
 
   .slick-prev {
     .@{carousel-prefix-cls}-rtl & {
```
WwwHhhYran: I found that when the layout direction is changed to RTL, the prev and next arrows do not adjust their positions. Maybe we should not use `.@{carousel-prefix-cls}-rtl & {}`, but use the following method instead, WDYT?
```less
&-rlt {
	// ...

	.slick-prev {
		// ...

		&::after {
			// ...
		}
	}
}
```
截屏2025-08-27 11 31 13

components/carousel/style/rtl.less
```diff
@@ -17,22 +17,24 @@
...
       right: auto;
-      left: -25px;
+      left: @carousel-arrow-offset;
 
-      &::before {
-        content: '←';
+      &::after {
+        // In RTL, next points to the left
+        transform: rotate(135deg);
```
WwwHhhYran:
```suggestion
        transform: rotate(-45deg);
```
WDYT ?

components/carousel/style/rtl.less
```diff
@@ -17,22 +17,24 @@
...
     .@{carousel-prefix-cls}-rtl & {
-      right: -25px;
+      right: @carousel-arrow-offset;
       left: auto;
 
-      &::before {
-        content: '→';
+      &::after {
+        // In RTL, previous points to the right
+        transform: rotate(-45deg);
```
WwwHhhYran:
```suggestion
        transform: rotate(135deg);
```
WDYT ?

components/carousel/style/rtl.less
Nicoss54: Ohhhh I think I have understood :)

Nicoss54: @WwwHhhYran after checking the documentation of antd, the behavior we have currently without apply suggest change is the same as the official documentation

I suggest to keep it like that to have the same behavior  as the officical documentation 

WDYT ? 

here is a print screen of the official ant design documentation:

Screenshot 2025-08-27 at 15 25 46

WwwHhhYran: @Nicoss54 Yes, I have also checked the antd documentation, but I think it's counterintuitive. I'm not even sure if it's a design specification or a bug, so I will try to confirm this with the antd community

Nicoss54: I agree :) but even if it’s rtl arrow are well designed but I thinks it’s the behavior that changed. Previous become next and next become previous ?

WDYT?

WwwHhhYran: > I agree :) but even if it’s rtl arrow are well designed but I thinks it’s the behavior that changed. Previous become next and next become previous ?
> 
> WDYT?

I think we use CSS to move the position of arrows and rotate them just to adapt to the layout direction, but the behavior of pre/next arrow hasn't changed: Clicking on the pre arrow will still switch to the previous content, and similarly, clicking on the next arrow will still switch to the next content. 

I'm not sure if my statement is clear, so I'll use an example to illustrate:

When we switch direction to RTL, the next arrow is now on the right and pointing to the right, and after adjustment, the next arrow will be on the left and pointing to the left (Just like the suggested change below https://github.com/NG-ZORRO/ng-zorro-antd/pull/9355#discussion_r2303064868). At this point, clicking on the next arrow will switch to the second page, and the dots below will also highlight the second one (located to the left of the first dot). This behavior is the same as when the direction is ltr.

Nicoss54: yeas make sense undestand now, this could be fix without changing the css, but just have a condition on the button or of course we could just use css with absolut position

I'm not sure about which solution is cleaner WDYT ?

WwwHhhYran: I think using CSS is better. On the one hand, you have already completed the `slick-next` and `slick-prev` CSS classes in the RTL direction. On the other hand, I prefer to use CSS for style issues rather than modifying component logic

WDTY? :)

Nicoss54: You’re agree. Let’s do that. When you have the feedback of the community I will update the code :) 

are you agree with this ? :)

Nicoss54: @WwwHhhYran do you have any news ?

WwwHhhYran: @Nicoss54 I'm sorry, I was on vacation a few days ago and didn't reply promptly. The antd community has confirmed that this is a bug https://github.com/ant-design/ant-design/issues/54781

Nicoss54: Hope you have some rest :) so let’s fix it. I will finished the dev today :)

Nicoss54: @WwwHhhYran it's done, arrow are reverse correctly in rtl mode :)

components/carousel/carousel.spec.ts
```diff
@@ -508,6 +582,45 @@ export class NzTestCarouselActiveIndexComponent {
...
+})
+export class NzTestCarouselArrowsComponent {
+  @ViewChild(NzCarouselComponent, { static: true }) nzCarouselComponent!: NzCarouselComponent;
+  array = [1, 2, 3, 4];
+  loop = true;
+}
+
+@Component({
+  selector: 'nz-test-carousel-no-swipe',
+  standalone: true,
```
Laffery: same here

components/carousel/carousel.spec.ts
```diff
@@ -508,6 +582,45 @@ export class NzTestCarouselActiveIndexComponent {
   }
 }
 
+@Component({
+  selector: 'nz-test-carousel-arrows',
+  standalone: true,
```
Laffery: please remove redundant `standalone: true`

Nicoss54:
@Laffery your changes requested are done :)

Laffery (approved):
LGTM

PR #9360, opened 2025-08-28
refactor(module:*): enhance popup arrow positioning and styles

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

The new arrow works well, take a look on the following screenshots:

### Tooltip

image

image

> By the way, why `RB` but the arrow points to the rightTop, I think this is a bug.

image

### Popover

image

### Dropdown

image

- [ ] Yes
- [x] No

## Other information

#9324 depends on this, and this PR can solve the related issue in #9333

### DIFF
--- a/components/core/overlay/overlay-position.ts
+++ b/components/core/overlay/overlay-position.ts
@@ -42,7 +42,36 @@ export type POSITION_TYPE_HORIZONTAL = Extract<
   'bottomLeft' | 'bottomCenter' | 'bottomRight' | 'topLeft' | 'topCenter' | 'topRight'
 >;
 
-export const DEFAULT_TOOLTIP_POSITIONS = [POSITION_MAP.top, POSITION_MAP.right, POSITION_MAP.bottom, POSITION_MAP.left];
+/**
+ * @internal
+ * @param offset offset in pixels which should not be less than 0.
+ * The default value is `12`, which means `(arrow-size / 2) + 4`
+ */
+const positionOffsetMapFactory = (offset: number = 12): Record<string, [number, number]> => ({
+  top: [0, -offset],
+  topCenter: [0, -offset],
+  topLeft: [0, -offset],
+  topRight: [0, -offset],
+  right: [offset, 0],
+  rightTop: [offset, 0],
+  rightBottom: [offset, 0],
+  bottom: [0, offset],
+  bottomCenter: [0, offset],
+  bottomLeft: [0, offset],
+  bottomRight: [0, offset],
+  left: [-offset, 0],
+  leftTop: [-offset, 0],
+  leftBottom: [-offset, 0]
+});
+
+export const TOOLTIP_OFFSET_MAP = positionOffsetMapFactory();
+
+export const DEFAULT_TOOLTIP_POSITIONS = [
+  setConnectedPositionOffset(POSITION_MAP.top, TOOLTIP_OFFSET_MAP.top),
+  setConnectedPositionOffset(POSITION_MAP.right, TOOLTIP_OFFSET_MAP.right),
+  setConnectedPositionOffset(POSITION_MAP.bottom, TOOLTIP_OFFSET_MAP.bottom),
+  setConnectedPositionOffset(POSITION_MAP.left, TOOLTIP_OFFSET_MAP.left)
+];
 
 export const DEFAULT_CASCADER_POSITIONS = [
   POSITION_MAP.bottomLeft,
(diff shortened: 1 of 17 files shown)


Nicoss54 (changes requested):
@Laffery could you make the pipeline pass ?

Laffery:
> @Laffery could you make the pipeline pass ?

CI passed :)

Nicoss54:
> > @Laffery could you make the pipeline pass ?
> 
> CI passed :)

you rock, i take the review now :) !

Nicoss54 (approved):
Basically LGTM, seems good to me after testing but i'm not an expert with Less

@HyperLife1119 could you review too please ?

lppedd:
@Laffery it looks like this PR broke `@popover-distance`.  
There is no way to control the distance between the arrow and the origin.

20.4.0:

image

20.1.2

image

Laffery:
> @Laffery it looks like this PR broke `@popover-distance`. There is no way to control the distance between the arrow and the origin.

Confirmed.
In this PR we align the CSS implementation of placement-arrow with antd@5.

In the past (< 20.2), zorro used `padding` and you can use `@popover-distance` to control the distance between the popover panel and the trigger element.

image

Since this PR, we are now using `CdkConnectedOverlay` to set the distance, therefore `@popover-distance` no longer works for this usage.

components/core/overlay/overlay-position.ts
```diff
@@ -42,7 +42,36 @@ export type POSITION_TYPE_HORIZONTAL = Extract<
   'bottomLeft' | 'bottomCenter' | 'bottomRight' | 'topLeft' | 'topCenter' | 'topRight'
 >;
 
-export const DEFAULT_TOOLTIP_POSITIONS = [POSITION_MAP.top, POSITION_MAP.right, POSITION_MAP.bottom, POSITION_MAP.left];
+/**
+ * @internal
+ * @param offset offset in pixels which should not be less than 0.
+ * The default value is `12`, which means `(arrow-size / 2) + 4`
+ */
+const positionOffsetMapFactory = (offset: number = 12): Record<string, [number, number]> => ({
```
Laffery, line 50: @lppedd look at here

lppedd: @Laffery thanks for the pointer! I'm not against aligning with AntDv5 and I agree that using CDK is better.

However I think an input to customize the distance would be better, simply because it improves usability and appearence (see my screenshot) for customized themes. The input value should then be summed to the arrow size.

I guess a temporary workaround could be changing the values of `TOOLTIP_OFFSET_MAP` at app start.

Laffery: Absolutely right, I will add an input to improve it

lppedd: @Laffery thank you, appreciate! My idea would be input + global config (let's say we want to align all tooltips/popovers), but it is up to you and how much time you want to invest into it, no worries.

lppedd: If anyone else needs to set a custom offset, this does the trick in the meantime.

```ts
provideAppInitializer(() => {
  const offset = 8; // arrow-size / 2 -> removed the additional 4px
  TOOLTIP_OFFSET_MAP["top"][1] = -offset;
  TOOLTIP_OFFSET_MAP["topCenter"][1] = -offset;
  TOOLTIP_OFFSET_MAP["topLeft"][1] = -offset;
  TOOLTIP_OFFSET_MAP["topRight"][1] = -offset;
  TOOLTIP_OFFSET_MAP["right"][0] = offset;
  TOOLTIP_OFFSET_MAP["rightTop"][0] = offset;
  TOOLTIP_OFFSET_MAP["rightBottom"][0] = offset;
  TOOLTIP_OFFSET_MAP["bottom"][1] = offset;
  TOOLTIP_OFFSET_MAP["bottomCenter"][1] = offset;
  TOOLTIP_OFFSET_MAP["bottomLeft"][1] = offset;
  TOOLTIP_OFFSET_MAP["bottomRight"][1] = offset;
  TOOLTIP_OFFSET_MAP["left"][0] = -offset;
  TOOLTIP_OFFSET_MAP["leftTop"][0] = -offset;
  TOOLTIP_OFFSET_MAP["leftBottom"][0] = -offset;
}),
```

Edit: it seems that the `left*` and `right*` offsets are not respected when the position is calculated automatically depending on the available space.

PR #9371, opened 2025-08-31
refactor(module:mention): remove mention service to use signal

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, the mention component uses the mention service only to be notified once the MentionTrigger directive has correctly registered.

Today, with signals, this service is no longer needed; the simple combination of effect with the `contentChild` function is enough to accomplish this part.

## What is the new behavior?

This pull request introduces a refactor of the mention service to use signals. This refactor has no impact on the component's functionality, and no breaking changes are expected.

The modified tests were changed only to await `afterNextRender`, which replaced the `afterViewInit` hook in the MentionTrigger directive.

- [ ] Yes
- [X] No


components/mention/mention.component.ts
```diff
@@ -322,13 +327,15 @@ export class NzMentionComponent implements OnInit, AfterViewInit, OnChanges {
   }
 
   private handleClick(): void {
-    this.resetDropdown();
+    this.resetDropdown(false);
```
Nicoss54, line 326: This take me a long time to investigate after the refactor, but finally i found the problem :) 

The issue was that after the signal refactor, both click and input events were calling resetDropdown() with emit: true, causing nzOnSearchChange to be emitted twice instead of once.

In the original service-based implementation, there was likely a timing issue or subscription management that prevented the click event from properly emitting the search change event,
 
Making the test expect only 1 call instead of the functionally correct 2 calls.

From my point of view, when user click on the input, nzOnSearch do not have to emit, that why i had false here.

Nicoss54:
@Laffery mention service is now removed replaced by signals 🥳

components/mention/mention.component.ts
```diff
@@ -158,6 +156,8 @@ export class NzMentionComponent implements OnInit, AfterViewInit, OnChanges {
     }
   }
 
+  trigger = contentChild.required(NzMentionTriggerDirective);
```
Laffery:
```suggestion
 readonly trigger = contentChild.required(NzMentionTriggerDirective);
```

components/mention/mention.component.ts
```diff
@@ -445,7 +446,7 @@ export class NzMentionComponent implements OnInit, AfterViewInit, OnChanges {
       const clickTarget = event.target as HTMLElement;
       return (
         this.isOpen &&
-        clickTarget !== this.trigger.elementRef.nativeElement &&
+        clickTarget !== this.trigger().elementRef.nativeElement &&
```
Laffery:
```suggestion
        clickTarget !== this.triggerNativeElement &&
```

Laffery:
basicly LGTM, just some small changes

components/mention/mention.component.ts
Nicoss54: done

components/mention/mention.component.ts
Nicoss54: done

Nicoss54:
@Laffery your changed requested are implemented :)

Laffery (approved):
LGTM

PR #9374, opened 2025-09-01
feat(module:core): support `withConfigFactory` for `input` properties

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

`@WithConfig` have been widely used to decorate on `@Input` properties, but it is not compatible with the `input` properties.

In this PR, we introduce a new `withConfigFactory` function, which can be used to generate a `withConfig` function, and its logic is same with `@WithConfig`.

### Usage Example

Firstly you should create a `withConfig` function with the factory and specific a `_nzModuleName`.

Then, add `input` property (or transform from the previous `@Input` property).

The last step is to declare a inner (**private** or **protected**) state with the `withConfig` function.

```ts
const withConfig = withConfigFactory('button');

class ExampleComponent {
  readonly nzSize = input<NzButtonSize>('default');
  protected readonly size = withConfig('nzSize', this.nzSize);
}
```

**The `size` property carries the latest value merged from the user inputs and the global config, we should use it instead of `nzSize` in the component logic.**
 If the `nzSize` property is not assigned by user, it will try to load the latest global config and apply to `size`. If the `nzSize` is assigned or changed by user, it wll also apply to `size`. 

- [ ] Yes
- [x] No

### DIFF
--- a/components/core/config/config.service.ts
+++ b/components/core/config/config.service.ts
@@ -10,12 +10,15 @@ import {
   afterNextRender,
   assertInInjectionContext,
   inject,
-  InputSignal,
   Signal,
-  computed
+  computed,
+  WritableSignal,
+  signal,
+  InputSignalWithTransform,
+  InputSignal
 } from '@angular/core';
 import { SIGNAL } from '@angular/core/primitives/signals';
-import { takeUntilDestroyed, toSignal } from '@angular/core/rxjs-interop';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { Observable, Subject, Subscription } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
 
@@ -30,13 +33,18 @@ const isDefined = function (value?: NzSafeAny): boolean {
 
 const defaultPrefixCls = 'ant';
 
-// todo: use nested signal to refactor the whole config service
-
 @Injectable({
   providedIn: 'root'
 })
 export class NzConfigService {
   private configUpdated$ = new Subject<keyof NzConfig>();
+
+  /**
+   * Sharing config signals for all components, used for {@link withConfigFactory}
+   * @internal
+   * @todo use nested signal to refactor the whole config service
+   */
+  private readonly _configMap = new Map<NzConfigKey, WritableSignal<NzConfig[NzConfigKey]>>();
 
   /** Global config holding property. */
   private readonly config: NzConfig = inject(NZ_CONFIG, { optional: true }) || {};
@@ -50,19 +58,23 @@ export class NzConfigService {
     }
   }
 
+  private _getConfigValue<T extends NzConfigKey>(componentName: T): WritableSignal<NzConfig[T]> {
+    let configValue = this._configMap.get(componentName) as WritableSignal<NzConfig[T]>;
+    if (configValue) {
+      return configValue;
+    }
+
+    configValue = signal(this.config[componentName]);
+    this._configMap.set(componentName, configValue);
+    return configValue;
+  }
+
   getConfig(): NzConfig {
     return this.config;
   }
 
   getConfigForComponent<T extends NzConfigKey>(componentName: T): NzConfig[T] {
     return this.config[componentName];
-  }
-
-  getConfigForComponentChange<T extends NzConfigKey>(componentName: T): Observable<NzConfig[T]> {
-    return this.configUpdated$.pipe(
-      filter(n => n === componentName),
-      map(() => this.getConfigForComponent(componentName))
-    );
   }
 
   getConfigChangeEventForComponent(componentName: NzConfigKey): Observable<void> {
@@ -74,6 +86,7 @@ export class NzConfigService {
 
   set<T extends NzConfigKey>(componentName: T, value: NzConfig[T]): void {
     this.config[componentName] = { ...this.config[componentName], ...value };
+    this._configMap.get(componentName)?.set(this.config[componentName]);
     if (componentName === 'theme' && this.config.theme) {
       registerTheme(this.getConfig().prefixCls?.prefixCls || defaultPrefixCls, this.config.theme, this.cspNonce);
     }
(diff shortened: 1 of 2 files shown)


HyperLife1119:
Using a decorator here introduces two DX issues:
- You have to write the type repeatedly
- You have to add a non-null assertion

```ts
@WithConfigSignal('nzSize')
protected readonly size!: WritableSignal<NzButtonSize>;
                       ^
                       In strict mode, a non-null assertion must be added.
```

I prefer an API like this:

```ts
@Component({
  providers: [
    { provide: NZ_MOUDLE_NAME, useValue: 'button' }
    // { provide: NZ_MOUDLE, useValue: { name: 'button' } }
  ]
})
class ExampleComponent {
  readonly nzSize = input<NzButtonSize>('default');

  readonly size = withConfig('nzSize', this.nzSize)
}
```

Or:

```ts
@NzModule('button') // maybe `@NzModule({ name: 'button', ...more })` ?
@Component(...)
class ExampleComponent {
  readonly size = input<NzButtonSize>('default', { alias: 'nzSize' });

  @WithConfig() readonly nzSize = linkedSignal(this.size)
}
```


Let me know what you think :)

Laffery:
> Let me know what you think :)

I prefer `withConfig` API, which follows Angular's trending. The only thing I concerned is the 1st argument cannot be auto-complete by ts

Laffery:
I update the implementation, which is much more TypeScript friendly, and we can reuse the inner signal for all of the inputs over the component


```ts
export function withConfigFactory<K extends NzConfigKey>(configKey: K) {
  const nzConfigService = inject(NzConfigService);

  const configValueSignal = toSignal(nzConfigService.getConfigChangeEventForComponent(configKey), {
    initialValue: nzConfigService.getConfigForComponent(configKey)
  });

  return <N extends keyof NonNullable<NzConfig[K]>, T extends NonNullable<NzConfig[K]>[N]>(
    name: N,
    inputSignal: InputSignal<T>
  ): Signal<T> => {
    const originalValue = inputSignal();

    return computed<T>(() => {
      const configValue = configValueSignal()?.[name] as T | undefined;
      const inputValue = inputSignal();
      // if the version of the inputSignal is 0 or the inputValue is undefined, we consider it as not assigned by user
      const assignedByUser = inputSignal[SIGNAL].version > 0 && isDefined(inputValue);

      if (assignedByUser) {
        return inputValue;
      } else if (configValue !== undefined) {
        return configValue;
      }

      return originalValue;
    });
  };
}
```


usage 

```ts
@Componnet({ 
  // ...
})
class ExampleComponent {
   private readonly withConfig = withConfigFactory('button');

   readonly nzSize = input<NzButtonSize>('default');
   readonly nzFoo = input<Foo>('foo');

   protected readonly innerSize = this.withConfig('nzSize', this.nzSize);
   protected readonly innerFoo = this.withConfig('nzFoo', this.nzFoo);
}
```

@HyperLife1119 WDYT?

HyperLife1119:
Looks good, but I think withConfig does not need to be a class member, we can put it outside:

```ts
const withConfig = withConfigFactory('button');

@Componnet({ 
  // ...
})
class ExampleComponent {
   readonly nzSize = input<NzButtonSize>('default');
   readonly nzFoo = input<Foo>('foo');

   protected readonly innerSize = withConfig('nzSize', this.nzSize);
   protected readonly innerFoo = withConfig('nzFoo', this.nzFoo);
}
```

This way we can get shorter code and better code compression.

The disadvantage is that it will inject and create toSignal every time it is called. What do you think?

Laffery:
You're right. 

I think the sharing `configValueSignal` should be implemented in the `NzConfigService` itself, so that we need not to maintain it in the factory func.
In the feature, I believe we can implement the sharing config value signals with `Deep Signal` (aka nested signal)

components/core/config/config.service.ts
```diff
@@ -54,6 +86,7 @@ export class NzConfigService {
 
   set<T extends NzConfigKey>(componentName: T, value: NzConfig[T]): void {
     this.config[componentName] = { ...this.config[componentName], ...value };
+    this._configMap.get(componentName)?.set(this.config[componentName]);
```
WwwHhhYran, line 89: 这里 set 的值应该和上面 `this.config[componentName]` 的赋值保持一致吧？否则`configMap` 和 `config` 中的值可能会出现不一致。而且 `_getConfigValue` 中的 else 分支也会从 `config` 取值为 `_configMap` 赋值，所以我理解这俩实际上应该是保持一致的。甚至 `_configMap` 是冗余的，为什么不直接把`_getConfigValue` 逻辑迁移到 `withConfigFactory` 方法中呢？

Laffery: > 这里 set 的值应该和上面 this.config[componentName] 的赋值保持一致吧？否则configMap 和 config 中的值可能会出现不一致。

这里没太懂，L89 `this.config[componentName]` 就是 L88 赋值后的值，什么情况会出现不一致

> 而且 _getConfigValue 中的 else 分支也会从 config 取值为 _configMap 赋值，所以我理解这俩实际上应该是保持一致的。甚至 _configMap 是冗余的，为什么不直接把 _getConfigValue 逻辑迁移到 withConfigFactory 方法中呢？

确实可以把 _getConfigValue 逻辑迁移到 withConfigFactory 方法中，但这样你会在每一次调用 withConfigFactory 都会创建一个用来存储该组件配置的 signal。

`_configMap` 算是一个优化，用来共享 ConfigValue Signal。
无论是在 A 组件中的 a b 参数，还是 B 组件中的 a 参数，都是共享一个组件模块下的 ConfigValue Signal，这样至多有 n 个 signal（n 为 zorro 组件数）

WwwHhhYran: get，不一致的问题是我看错了 :)

WwwHhhYran (approved):

components/core/config/config.service.ts
Laffery, line 89: By the way, the reason why I not use `_getConfigValue` to retrive the Config Value Signal is that there may be no component using it. 
It will not create the specific signal until some component use it.

HyperLife1119 (approved):
LGTM

PR #9377, opened 2025-09-02
feat(module:mention): support nzClear

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Mention component now support nzClear icon

- [ ] Yes
- [X] No

### DIFF
--- a/components/mention/mention-trigger.ts
+++ b/components/mention/mention-trigger.ts
@@ -114,8 +114,9 @@ export class NzMentionTriggerDirective implements ControlValueAccessor {
   }
 
   writeValue(value: string): void {
-    this.value.set(value || '');
-    this.elementRef.nativeElement.value = this.value();
+    const parsedValue = typeof value === 'string' ? value : '';
+    this.value.set(parsedValue);
+    this.elementRef.nativeElement.value = parsedValue;
   }
 
   onChange: OnChangeType = () => {};


components/mention/mention.component.ts
```diff
@@ -118,13 +120,16 @@ export type MentionPlacement = 'top' | 'bottom';
     @if (hasFeedback && !!status) {
       <nz-form-item-feedback-icon class="ant-mentions-suffix" [status]="status" />
     }
+    @if (nzAllowClear && hasValue()) {
+      <nz-select-clear class="ant-mentions-suffix" [clearIcon]="nzClearIcon" (clear)="clear()" />
```
HyperLife1119: Image

Due to the different DOM structure, nz-select-clear cannot be used directly.

components/mention/mention.component.ts
```diff
@@ -168,6 +176,9 @@ export class NzMentionComponent implements OnInit, AfterViewInit, OnChanges {
   statusCls: NgClassInterface = {};
   status: NzValidateStatus = '';
   hasFeedback: boolean = false;
+  hasValue = computed(() => {
```
HyperLife1119:
```suggestion
  readonly hasValue = computed(() => {
```

components/mention/mention-trigger.ts
```diff
@@ -60,7 +61,8 @@ export class NzMentionTriggerDirective implements ControlValueAccessor {
   @Output() readonly onKeydown = new EventEmitter<KeyboardEvent>();
   // eslint-disable-next-line @angular-eslint/no-output-on-prefix
   @Output() readonly onClick = new EventEmitter<MouseEvent>();
-  value?: string;
+
+  value = signal<string>('');
```
HyperLife1119:
```suggestion
  readonly value = signal<string>('');
```

components/mention/mention-trigger.ts
```diff
@@ -100,16 +102,18 @@ export class NzMentionTriggerDirective implements ControlValueAccessor {
...
 
   writeValue(value: string): void {
-    this.value = value;
-    if (typeof value === 'string') {
-      this.elementRef.nativeElement.value = value;
-    } else {
-      this.elementRef.nativeElement.value = '';
-    }
+    this.value.set(value || '');
+    this.elementRef.nativeElement.value = this.value();
```
HyperLife1119: Can we use host bindings?

```ts
host: {
  '[value]': 'value()'
}
```

components/mention/mention.component.ts
Nicoss54: done :)

components/mention/mention.component.ts
Nicoss54: done :)

components/mention/mention-trigger.ts
Nicoss54: done :)

components/mention/mention-trigger.ts
Nicoss54: Yeah after investiging a little bit it seems possible. Making this changes introduce the possibility to clean and refactor the code like this one

```typescript
insertMention(mention: Mention): void {
    const value: string = this.elementRef.nativeElement.value; 
    const insertValue = `${mention.mention}${NZ_MENTION_CONFIG.split}`;
    const newValue = [
      value.slice(0, mention.startPos + 1),
      insertValue,
      value.slice(mention.endPos, value.length)
    ].join('');
    this.elementRef.nativeElement.value = newValue; // not necessary anymore it's always sync
    this.focus(mention.startPos + insertValue.length + 1);
    this.onChange(newValue);
    this.value.set(newValue);
  }
```


I could work on this just after if you want with great pleasure. But the refactor from my point of view have no concerns with this feature WDYT ?

HyperLife1119: OK, let's do this in another PR :)

Nicoss54:
@HyperLife1119 i have just updated the PR to avoid recreating the weel :) now it’s good :) thanks for your patience and review ! :)

HyperLife1119:
Oh no, can you undo it? I don't recommend using nz-select-clear here because it's not abstract enough :(

Nicoss54:
> Oh no, can you undo it? I don't recommend using nz-select-clear here because it's not abstract enough :(

Of course :)

Nicoss54:
> > Oh no, can you undo it? I don't recommend using nz-select-clear here because it's not abstract enough :(
> 
> Of course :)
 
@HyperLife1119 it’s revert :)

components/mention/demo/clear.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { NzInputModule } from 'ng-zorro-antd/input';
+import { NzMentionModule } from 'ng-zorro-antd/mention';
+
+@Component({
+  selector: 'nz-demo-mention-clear',
+  imports: [FormsModule, NzInputModule, NzMentionModule],
+  template: `
+    <nz-mention [nzSuggestions]="suggestions" (nzOnSelect)="onSelect($event)" nzAllowClear (nzOnClear)="onClear()">
+      <textarea rows="1" placeholder="input here" nzMentionTrigger nz-input [(ngModel)]="inputValue"></textarea>
```
Laffery, line 12: would you like to use self closed tag here?

components/mention/demo/clear.ts
```diff
@@ -0,0 +1,32 @@
...
+    </nz-mention>
+    <br />
+    <br />
+    <nz-mention [nzSuggestions]="suggestions" (nzOnSelect)="onSelect($event)" nzAllowClear (nzOnClear)="onClear()">
+      <textarea rows="3" placeholder="input here" nzMentionTrigger nz-input [(ngModel)]="inputValue"></textarea>
```
Laffery, line 17: same here

components/mention/mention.spec.ts
```diff
@@ -699,6 +781,33 @@ class NzTestMentionInFormComponent {
...
+@Component({
+  imports: [FormsModule, NzInputModule, NzMentionModule],
+  template: `
+    <nz-mention
+      [nzSuggestions]="suggestions"
+      [nzAllowClear]="allowClear"
+      [nzClearIcon]="useCustomClearIcon ? clearIconTemplate : null"
+      (nzOnClear)="onClear()"
+    >
+      <textarea nz-input [nzAutosize]="{ minRows: 4, maxRows: 4 }" [(ngModel)]="inputValue" nzMentionTrigger></textarea>
```
Laffery, line 900: same here

components/mention/demo/clear.ts
HyperLife1119: Textarea is a native html element and does not support self-closing.

Nicoss54:
@HyperLife1119 could I merge it ?

components/mention/mention-trigger.ts
```diff
@@ -103,16 +104,18 @@ export class NzMentionTriggerDirective implements ControlValueAccessor {
...
   }
 
   writeValue(value: string): void {
-    this.value = value;
-    if (typeof value === 'string') {
-      this.elementRef.nativeElement.value = value;
-    } else {
-      this.elementRef.nativeElement.value = '';
-    }
+    this.value.set(value || '');
```
Laffery: What if the user write a non-string value? There seems to be a different behavior with this PR

Nicoss54: But the type of value in the previous pr was a string. And it could be at the end only a string  cause the directive is apply on an input or a text area so when we enter in the writeValue hook the value can only be a string or maybe I miss something in my reasoning ?

Laffery: The NzMentionTriggerDirective is deigned to used on an `input` or a `textarea` element, which value should be string you're right.

However, there is no type check with `ngModel`. for example:

```ts
@Component({
  imports: [FormsModule, NzMentionTriggerDirective],
  template: `
   <nz-mention [nzSuggestions]="suggestions" (nzOnSelect)="onSelect($event)">
      <textarea
        rows="1"
        placeholder="input here"
        nzMentionTrigger
        nz-input
        [(ngModel)]="value"
        (ngModelChange)="onChange($event)"
      ></textarea>
    </nz-mention>
  `
})
export class MentionTestComponent {
  value = model<object>({});
}
```

The result is

image

The previous result is

image

Nicoss54: Ohhh I totally missed the use case of NgModel, that's true. I will revert this change and then I think we will be good

Nicoss54:
@Laffery change are done, I revert to the previous behavior

Laffery (approved):
LGTM

PR #9563, opened 2025-11-27
test(module:avatar): update tests to use zoneless mode

Nicoss54:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Avatar test still works in Zone mode

Issue Number: N/A

## What is the new behavior?

Avatar test now works i zoneless mode as the default mode in Angular 21

- [ ] Yes
- [X] No

### DIFF
--- a/components/avatar/avatar.spec.ts
+++ b/components/avatar/avatar.spec.ts
@@ -3,11 +3,11 @@
  * found in the LICENSE file at https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/LICENSE
  */
 
-import { Component, DebugElement, provideZoneChangeDetection, ViewChild } from '@angular/core';
-import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
+import { Component, DebugElement, ViewChild } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
-import { createFakeEvent } from 'ng-zorro-antd/core/testing';
+import { createFakeEvent, updateNonSignalsInput } from 'ng-zorro-antd/core/testing';
 import { NzSafeAny, NzShapeSCType, NzSizeLDSType } from 'ng-zorro-antd/core/types';
 import { provideNzIconsTesting } from 'ng-zorro-antd/icon/testing';
 
@@ -31,13 +31,14 @@ function getType(dl: DebugElement): string {
 
 describe('avatar group', () => {
   let fixture: ComponentFixture<TestAvatarGroupComponent>;
-  beforeEach(() => {
+  beforeEach(async () => {
+    TestBed.configureTestingModule({});
     fixture = TestBed.createComponent(TestAvatarGroupComponent);
-    fixture.detectChanges();
+    fixture.autoDetectChanges();
   });
 
-  it('should avatar group work', () => {
-    fixture.detectChanges();
+  it('should avatar group work', async () => {
+    await fixture.whenStable();
     const avatarGroup = fixture.debugElement.query(By.directive(NzAvatarGroupComponent));
     expect(avatarGroup.nativeElement.classList).toContain('ant-avatar-group');
   });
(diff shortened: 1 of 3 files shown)


components/avatar/avatar.spec.ts
```diff
@@ -31,13 +31,17 @@ function getType(dl: DebugElement): string {
 
 describe('avatar group', () => {
   let fixture: ComponentFixture<TestAvatarGroupComponent>;
-  beforeEach(() => {
+  beforeEach(async () => {
+    TestBed.configureTestingModule({
```
Laffery: remove empty providers

components/avatar/avatar.spec.ts
```diff
@@ -31,13 +31,17 @@ function getType(dl: DebugElement): string {
...
     fixture = TestBed.createComponent(TestAvatarGroupComponent);
-    fixture.detectChanges();
+    fixture.autoDetectChanges();
+    await fixture.whenStable();
   });
 
-  it('should avatar group work', () => {
-    fixture.detectChanges();
+  it('should avatar group work', async () => {
+    await fixture.whenStable();
```
Laffery, line 41: `await fixture.whenStable` has been executed at L40

From my perspective, it's better to do it in each case rather than in `beforeEach`, because  it may update some inputs before `await fixture.whenStable` in most cases.

components/core/testing/zoneless-helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import type { ComponentFixture } from '@angular/core/testing';
+
+export function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+export async function updateNonSignalsInput<T>(fixture: ComponentFixture<T>, ms: number = 0): Promise<void> {
+  fixture.changeDetectorRef.markForCheck();
+  await sleep(ms);
```
Laffery: If `ms` is 0, it could have 2 semantics:

- sleep 0ms
- skip sleeping

is there an option to skip sleeping?

components/avatar/avatar.spec.ts
```diff
@@ -52,164 +56,153 @@ describe('avatar', () => {
...
       event.preventDefault();
       expect(getType(dl)).toBe('image');
       expect(context.comp.hasSrc).toBe(true);
       // Manually dispatch error.
       context.nzSrc = 'Invalid image src';
+      await updateNonSignalsInput(fixture);
       context.comp.imgError(event);
-      tick();
-      fixture.detectChanges();
+      await updateNonSignalsInput(fixture);
```
Laffery: `imgError` does not change any inputs

Laffery (changes requested):

components/core/testing/zoneless-helpers.ts
Nicoss54: No for the moment there is no option to skip the sleep part. Maybe we could have a third argument in this function like skipSleep which by default is set to false?

components/avatar/avatar.spec.ts
Nicoss54: You've right. I didn't think of that ! Thanks for remember me that ! :)

components/core/testing/zoneless-helpers.ts
WwwHhhYran: I have a question 🤔 

Why not separate `sleep` and `updateNonSignalsInput` ? I dont think there is any connection between this two, they are independent. We can let other developers combine them for use or use them independently.

Nicoss54: true, i separate the two function :)

components/avatar/avatar.spec.ts
Nicoss54: done

components/avatar/avatar.spec.ts
Nicoss54: done

Nicoss54:
@Laffery @WwwHhhYran i implement the change you asked, now avatar test run on zoneless mode. If you're agree with this PR i could continue the migration

components/core/testing/zoneless-helpers.ts
Laffery: looks like the two function have not been separated yet.

> Maybe we could have a third argument in this function like skipSleep which by default is set to false?

What about using `undefined` to indicate skipping `sleep`, WDYT?

```ts
export async function updateNonSignalsInput<T>(fixture: ComponentFixture<T>, ms?: number): Promise<void> {
  fixture.changeDetectorRef.markForCheck();
  if (typeof ms === 'number') {await sleep(ms);}
  await fixture.whenStable();
}
```

In this way, if developer not gonna `sleep`, not pass the second argument.

Nicoss54: It looks like I forgot to push ...  make it as soon as possible

Nicoss54: @Laffery all changes are impelmented and push :)

Laffery (approved):
LGTM

WwwHhhYran (approved):
LGTM

PR #9612, opened 2025-12-21
fix(module:carousel): restore responsive height on window resize

### LINKED ISSUE #8572: nz-carousel height no longer responsive
### Reproduction link
[https://stackblitz.com/edit/ng-zorro-antd-ivy-w2drwm?file=src%2Fapp%2Fapp.component.html](https://stackblitz.com/edit/ng-zorro-antd-ivy-w2drwm?file=src%2Fapp%2Fapp.component.html)

### Steps to reproduce
On load the issue is already present, the "previous" and "next" buttons get pushed further down than should be. (As seen by the 2px wide red border.

When pressing on the "toggle" button situated at the top of the page, we see a shift of the buttons when the nz-carousel component gets destroyed.
Pressing it once more, makes it reappear in the correct dimensions.

But in either case, vertical responsiveness is not working and the nz-carousel does not change height.

### What is expected?
nz-carousel 's vertical responsiveness is intact and follows along when resizing the window 

nz-carousel exceeds the allotted space, causing elements to not be situated correctly.

| Environment | Info |
|---|---|
| ng-zorro-antd | 17.4.1 |
| Browser | Chrome |

---
I've been struggling to get this issue fixed for a few days now and am out of ideas or way to fix it, I apologise for any inconvenience

ashraf7hossain:
fix(carousel): restore responsive height on window resize

## PR Checklist
- [ ] Add unit tests for the fix
- [ ] Tested in latest Chrome/Safari/Firefox
- [ ] Updated documentation (if needed)
- [ ] Followed coding style

## Description
The nz-carousel component no longer adjusts its height responsively when the window is resized, causing overflow and mispositioned navigation buttons (as reported in #8572).

This PR fixes the issue by [briefly describe your changes, e.g., "adding a ResizeObserver to dynamically update the carousel height based on content"].

## Reproduction
Reproduced on ng-zorro-antd v17.4.1 (see original StackBlitz: https://stackblitz.com/edit/ng-zorro-antd-ivy-w2drwm?file=src%2Fapp%2Fapp.component.html).

## Before/After
- Before: Height fixed, overflows on resize.
- After: Height adjusts dynamically.

Fixes #8572

Thanks for reviewing!


components/carousel/style/index.less
```diff
@@ -67,8 +67,11 @@
   .slick-slider .slick-list {
     transform: translate3d(0, 0, 0);
     touch-action: pan-y;
+    height: auto !important;
```
Nicoss54: nitpick(if-minor): is it possible to avoid having the important

components/carousel/style/index.less
```diff
@@ -67,8 +67,11 @@
...
   }
 
+
```
Nicoss54: nitpick(if-minor): is it possible to format the file correctly ?

components/carousel/style/index.less
ashraf7hossain: I removed !important from my adjustment.

components/carousel/style/index.less
```diff
@@ -67,6 +67,7 @@
   .slick-slider .slick-list {
     transform: translate3d(0, 0, 0);
     touch-action: pan-y;
+    height: auto;
```
Laffery: This style doesn't exist in antd. Could you put it in `patch.less`?

Issue #9692, opened 2026-02-13
Docs/demo: Date Picker popup spawns in incorrect position


lppedd:
### Reproduction link
https://ng.ant.design/components/date-picker/en#components-date-picker-demo-range-picker  

### Steps to reproduce
See https://ng.ant.design/components/date-picker/en#components-date-picker-demo-range-picker  
I cannot reproduce in the Stackblitz link tho.

<img src="https://github.com/user-attachments/assets/eb43c2cb-c47f-491d-b4ad-a6326a695766" width="700">

### What is expected?
The picker popup points to the Data Picker component.

### What is actually happening?
The picker popup points to the wrong place in the page.

| Environment | Info |
|---|---|
| ng-zorro-antd | 21.1.0 |
| Browser | Chrome 144.0.7559.133 |

Laffery:
I cannot reproduce it in https://ng.ant.design/components/date-picker/en#components-date-picker-demo-range-picker

lppedd:
@Laffery yup to reproduce you need to play with scroll position (basically where you're in the page).  
But here is an even simpler case I've just reproduced:

<img src="https://github.com/user-attachments/assets/89320769-b7bf-42b2-8daa-51054a0a421e" width="700">

lppedd:
@Laffery Also, try the following: open the **Switchable picker** example date picker, and start slowly scrolling up and down the page, multiple times. You'll see the popup drift away from the origin.

Laffery:
@lppedd I tried multiple times following your suggestions, but unfortunately I still couldn't reproduce it. If you have a stable way to reproduce this issue, please let me know.

lppedd:
@Laffery that's really odd, but thanks for trying! I can reproduce in multiple PCs (with different Chrome versions).

However, as I wrote earlier I can reproduce only in the Docs website, but not in StackBlitz.  
So it must be something related to the Docs website only. Maybe how the sections are split up.

But yeah I don't know how to prepare a proper reproducer yet. I'll let you know if I find a way.

https://github.com/user-attachments/assets/dd22cb69-56f9-49c1-8219-9d28f7f4b263

lppedd:
@Laffery AHHHH! Are you using the light/Default theme? If yes, switch to Dark!

In the Dark theme, you'll see that many overlay elements appear in the wrong place, not only the date ones.

Laffery:
[...]

PR #9622, opened 2025-12-26
refactor(module:date-picker): use native animations

Nicoss54:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Application (the showcase website) / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
- [ ] Yes
- [X] No

PR #9723, opened 2026-03-07
doc: append theme stylesheet to document head instead of body

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Application (the showcase website) / infrastructure changes

Issue Number: #9692

When switching to dark/compact/aliyun theme on the documentation site, the theme CSS `<link>` element was appended to `document.body` instead of `document.head`. This causes the browser to potentially create a new compositor layer on `body` (triggered by the active `transition: background` animation). As a result, the containing block for `position: fixed` elements changes temporarily, causing the CDK `.cdk-overlay-container` to miscalculate its position relative to the viewport — the Date Picker popup appears at the wrong location in dark mode.

## What is the new behavior?

The theme stylesheet `<link>` is now appended to `document.head` instead of `document.body`, ensuring the CDK overlay container's `position: fixed` positioning context always refers to the viewport correctly.

- [ ] Yes
- [x] No

## Other information

The issue could not be reproduced on Stackblitz because it does not use the documentation site's theme-switching mechanism (no `<link>` appended to `body`).

### DIFF
--- a/scripts/site/doc/app/app.component.ts
+++ b/scripts/site/doc/app/app.component.ts
@@ -150,7 +150,7 @@ export class AppComponent implements OnInit {
       style.rel = 'stylesheet';
       style.id = `site-theme-${theme}`;
       style.href = `assets/${theme}.css`;
-      document.body.append(style);
+      document.head.append(style);
 
       style.onload = () => {
         successLoaded();


Laffery:
## Additional Context: Why Did This Surface in v21?

The root cause (`document.body.append(style)`) has existed for a long time, but the issue only became visible with certain components (notably Date Picker) after v21. Here's why.

### Before v21 — Date Picker had an isolation layer

Prior to PR #9622 (`refactor(module:date-picker): use native animation`), the overlay template wrapped its content in an extra div with `position: relative`:

```html
<ng-template cdkConnectedOverlay
  [cdkConnectedOverlayTransformOriginOn]="'.ant-picker-wrapper'">

  <div class="ant-picker-wrapper"
       [@slideMotion]="'enter'"
       [style.position]="'relative'">   <!-- isolation layer -->
    <ng-container *ngTemplateOutlet="inlineMode"></ng-container>
  </div>

</ng-template>
```

The Angular `[@slideMotion]` animation applied `transform: scaleY()` only to this `position: relative` wrapper — never to the overlay pane itself. Because no `transform` existed on the pane level, even if `body` created a compositor layer during the theme transition, the `position: fixed` containing block was not affected.

### After v21 — isolation layer removed

PR #9622 migrated to native CSS animations and removed the wrapper entirely:

```html
<ng-template cdkConnectedOverlay
  cdkConnectedOverlayTransformOriginOn=".ant-picker-dropdown">

  <!-- wrapper div gone, CSS animation classes applied directly inside the pane -->
  <ng-container *ngTemplateOutlet="inlineMode"></ng-container>

</ng-template>
```

Now the CSS animation (e.g. `ant-slide-up-enter-active`) applies `transform` directly to the dropdown element inside the pane. When `dark.css` is appended to `body` and the `transition: background 1s` fires, the browser composites `body`, temporarily turning it into the containing block for `position: fixed` elements — and the overlay position calculation breaks.

### Why dropdown / select / tooltip were always affected

These components never had the `position: relative` wrapper pattern — their CSS animations have always targeted the dropdown element directly (`.ant-dropdown`, `.ant-select-dropdown`, `.ant-tooltip`). So they were subject to the same issue all along, just less likely to be noticed because:
- Tooltip is triggered on hover, rarely overlapping with a theme-switch moment
- The position offset scales with scroll depth, making it non-obvious near the top of the page

The fix (`document.head.append`) eliminates the `body` compositor layer entirely, resolving the issue for all CDK overlay-based components at once.

lppedd:
Just validated it using `gulp serve:site` and the issue is gone, thank you!

Not sure if Copilot/Gemini did the pre-v21 analysis above, but it seems good (I tend to be skeptical, but in theory it looks to be correct).

lppedd:
Btw, not strictly related but

https://github.com/NG-ZORRO/ng-zorro-antd/blob/f83af1e87a6eaab041c38f0ed7b26ff8d84568c4/scripts/gulp/util/task-helpers.ts#L28-L32

doesn't work on Windows. `spawn` does not allow `.cmd` files to run, unless `shell: true` is used.

Laffery:
> doesn't work on Windows. `spawn` does not allow `.cmd` files to run, unless `shell: true` is used.

I doesn't have a Windows machine to validate it, welcome to contribute :)

lppedd:
@Laffery no worries. I'll open a PR in the next week for that problem. I think this PR is good to go as it is!

PR #9773, opened 2026-05-01
fix(module:descriptions): use track functions in @for loops

### LINKED ISSUE #9587: Resizing window with nz-descriptions will trigger angular error
### Reproduction link
[https://stackblitz.com/edit/ng-zorro-antd-ivy-usrx3jis?file=src%2Fapp%2Fapp.component.ts](https://stackblitz.com/edit/ng-zorro-antd-ivy-usrx3jis?file=src%2Fapp%2Fapp.component.ts)

### Steps to reproduce
Resize the window

### What is expected?
Everything to be ok

NG0956: The configured tracking expression (track by identity) caused re-creation of the entire collection of size 5. This is an expensive operation requiring destruction and subsequent creation of DOM nodes, directives, components etc. Please review the "track expression" and make sure that it uniquely identifies items in a collection. Find more at https://v20.angular.dev/errors/NG0956

| Environment | Info |
|---|---|
| ng-zorro-antd | 20.4.3 |
| Browser | Chrome |

---
nz-descriptions seems to be using for loop but not using an unique tracking identity.

swseverance:
Replaces object references used directly as track expressions with protected track functions, resolving the NG0956 warning logged to the console.  This is the most direct approach to getting rid of the warning message. I suppose it is possible to refactor the component so that each row and item of `itemMatrix` has a unique identifier that the `track` function can use but I'm not sure such an approach is necessary here.

Resolves #9587

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Users are experiencing the NG0956 warning message being logged when using the `nz-descriptions` component.

Issue Number: 9587

## What is the new behavior?

There is no more warning message

- [ ] Yes
- [X] No

### DIFF
--- a/components/descriptions/descriptions.component.ts
+++ b/components/descriptions/descriptions.component.ts
@@ -69,9 +69,9 @@ const defaultColumnMap: Record<NzBreakpointEnum, number> = {
       <table>
         <tbody>
           @if (nzLayout === 'horizontal') {
-            @for (row of itemMatrix; track row; let i = $index) {
+            @for (row of itemMatrix; track trackItems(row); let i = $index) {
               <tr class="ant-descriptions-row">
-                @for (item of row; track item; let isLast = $last) {
+                @for (item of row; track trackItem(item); let isLast = $last) {
                   @if (!nzBordered) {
                     <td class="ant-descriptions-item" [colSpan]="item.span">
                       <div class="ant-descriptions-item-container">
@@ -102,9 +102,9 @@ const defaultColumnMap: Record<NzBreakpointEnum, number> = {
 
           @if (nzLayout === 'vertical') {
             @if (!nzBordered) {
-              @for (row of itemMatrix; track row; let i = $index) {
+              @for (row of itemMatrix; track trackItems(row); let i = $index) {
                 <tr class="ant-descriptions-row">
-                  @for (item of row; track item; let isLast = $last) {
+                  @for (item of row; track trackItem(item); let isLast = $last) {
                     <td class="ant-descriptions-item" [colSpan]="item.span">
                       <div class="ant-descriptions-item-container">
                         <span class="ant-descriptions-item-label" [class.ant-descriptions-item-no-colon]="!nzColon">
@@ -117,7 +117,7 @@ const defaultColumnMap: Record<NzBreakpointEnum, number> = {
                   }
                 </tr>
                 <tr class="ant-descriptions-row">
-                  @for (item of row; track item; let isLast = $last) {
+                  @for (item of row; track trackItem(item); let isLast = $last) {
                     <td class="ant-descriptions-item" [colSpan]="item.span">
                       <div class="ant-descriptions-item-container">
                         <span class="ant-descriptions-item-content">
@@ -129,9 +129,9 @@ const defaultColumnMap: Record<NzBreakpointEnum, number> = {
                 </tr>
               }
             } @else {
-              @for (row of itemMatrix; track row; let i = $index) {
+              @for (row of itemMatrix; track trackItems(row); let i = $index) {
                 <tr class="ant-descriptions-row">
-                  @for (item of row; track item; let isLast = $last) {
+                  @for (item of row; track trackItem(item); let isLast = $last) {
                     <td class="ant-descriptions-item-label" [colSpan]="item.span">
                       <ng-container *nzStringTemplateOutlet="item.title">
                         {{ item.title }}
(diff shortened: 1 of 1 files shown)


swseverance:
To provide a little bit more context, the angular code that logs the warning is the following:
```ts
    // Warn developers about situations where the entire collection was re-created as part of the
    // reconciliation pass. Note that this warning might be "overreacting" and report cases where
    // the collection re-creation is the intended behavior. Still, the assumption is that most of
    // the time it is undesired.
    if (
      ngDevMode &&
      metadata.trackByFn === ɵɵrepeaterTrackByIdentity &&
      liveCollection.operationsCounter?.wasReCreated(liveCollection.length) &&
      isViewExpensiveToRecreate(getExistingLViewFromLContainer(lContainer, 0))
    ) {
      const message = formatRuntimeError(
        RuntimeErrorCode.LOOP_TRACK_RECREATE,
        `The configured tracking expression (track by identity) caused re-creation of the entire collection of size ${liveCollection.length}. ` +
          'This is an expensive operation requiring destruction and subsequent creation of DOM nodes, directives, components etc. ' +
          'Please review the "track expression" and make sure that it uniquely identifies items in a collection.',
      );
      console.warn(message);
    }
```
Where `eerepeaterTrackByIdentity` is:
```ts
/**
 * A built-in trackBy function used for situations where users specified collection item reference
 * as a tracking expression. Having this function body in the runtime avoids unnecessary code
 * generation.
 *
 * @param index
 * @returns
 */
export function ɵɵrepeaterTrackByIdentity<T>(_: number, value: T) {
  return value;
}
```
Basically when you use your own trackByFn in the `@for` loop it overrides the default trackByFunction provided by angular. When that happens the check for if the warning should be logged exits early because the following evaluates to false:
```ts
metadata.trackByFn === ɵɵrepeaterTrackByIdentity
```

If you believe the issue should be resolved by introducing custom trackBy functions that use some sort of identifier instead of object identity I can close the PR. Thanks!

Nicoss54:
@swseverance 

Thanks for all the time you take for the investigration, but instead creating a function that return the item to use the overriding system from Angular



WDYT about just using index in this case.



Here is why i'm thinking with you solution:  track trackItem(item) evaluates to the same object reference as the original track item, so the tracking key is identical. That means NG0956 (duplicate keys) can still fire if itemMatrix contains the same row/item reference twice



If you'r agree with my idea i can make the PR, i already dit the code but forgot to push it

swseverance:
@Nicoss54 If I'm not mistaken the descriptions component basically subscribes to the changes Observable of `QueryList<NzDescriptionsItemComponent>` and then calls `prepareMatrix()`.  It seems like `prepareMatrix` iterates over the items and creates a new object each time. Here's an overly simplified excerpt of some of the code:
```ts
    for (let i = 0; i < length; i++) {
      const item = items[i];
      const { nzTitle: title, content, nzSpan: span } = item;
      currentRow.push({ title, content, span: column - (width - span) });
      flushRow();
```
So I don't think it will ever create the same object twice but I could be mistaken.

If your PR can use $index to solve the problem I totally agree it would be a better approach than the one I have provided. However I just wanted to point out that using $index can lead to some tricky edge cases. [I have prepared the following StackBlitz that uses a $index approach](https://stackblitz.com/edit/stackblitz-starters-jyvmdnvq?description=An%20angular-cli%20project%20based%20on%20@angular/animations,%20@angular/common,%20@angular/compiler,%20@angular/core,%20@angular/forms,%20@angular/platform-browser,%20@angular/platform-browser-dynamic,%20@angular/router,%20core-js,%20rxjs,%20tslib%20and%20zone.js&file=src%2Fmain.ts&template=node&title=Angular%20Starter).  

Please try the following:
1) Click the button next to "Alice" to open the card. You will see the "Bob" card is not expanded
2) Type "b" in the input (to filter only items that start with "b"
3) You will then see that the "Bob" card is expanded, which is not what a user would expect.

I'm only trying to illustrate that $index approach can be a little bit tricky at times. But, if you are able to make it work then I agree it would be a better solution!

Nicoss54:
@swseverance you're right each time it emits a value, prepareMatrix is called and we push the item in the matrix that is a new matrix. Miss that :)

About your exemple that's a classic issue I agree about track with $index 

Here is the explanation 

@For tracks items by their DOM position (index), not by identity. When the list shrinks due to filtering, Angular reuses existing DOM elements positionally — it doesn’t destroy Alice’s node and create a fresh one for Bob

But the $index for the description component context  work :) what I will make is to create a calculate identifier to avoid this kind of issue you report

swseverance:
@Nicoss54 a unique id sounds like a good approach. Thanks!

PR #9820, opened 2026-06-05
chore: bump angular to v22

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


components/icon/icon.directive.ts
```diff
@@ -34,81 +35,92 @@ import { NzIconPatchService, NzIconService } from './icon.service';
...
-    '[class]': `'anticon-' + type`,
-    '[class.anticon-spin]': `nzSpin || type === 'loading'`,
     role: 'img',
-    '[attr.aria-label]': 'type'
+    '[class]': `hostClass()`,
+    '[attr.aria-label]': 'nzType()'
   }
 })
-export class NzIconDirective extends IconDirective implements OnChanges, AfterContentChecked {
+export class NzIconDirective extends IconBase implements AfterContentChecked {
```
Laffery, line 43: see https://github.com/ant-design/ant-design-icons/pull/734

Nicoss54 (approved):
LGTM

Nicoss54 (changes requested):
Could you make the pipeline pass again please ?

Laffery:
> Could you make the pipeline pass again please ?

I'm on it, which is caused by synchronization delays on the npmmirror registry. I am manually triggering the sync for each package one by one.

Nicoss54:
> > Could you make the pipeline pass again please ?
> 
> 
> 
> I'm on it, which is caused by synchronization delays on the npmmirror registry. I am manually triggering the sync for each package one by one.

Got it !

Laffery:
CI npm run test Exit Issue - Root Cause Analysis

The CI pipeline hangs after npm run test completes because Angular 22's `@angular/build:karma`
builder leaves an esbuild service subprocess orphaned after single-run finishes.

**Root Cause**

After tests complete, a node + esbuild --service=0.28.0 --ping process remains alive, keeping the
Node event loop from exiting. This is tracked in [angular/angular-cli#33201](https://github.com/angular/angular-cli/issues/33201) (closed via #33267):

> For watch: false, BundlerContext.incremental is false, leading to a one-shot esbuild.build() without
    creating an esbuild.context(). ExecutionResult.dispose() only awaits
   this.#esbuildContext?.dispose(), which is undefined in this path, failing to tear down the esbuild
   service child. The builder never explicitly calls esbuild.stop().

**Fix Applied**

Added a custom Karma reporter (AngularBuildCleanupReporter) in karma.conf.js that calls
esbuild.stop() during the onExit phase to clean up the orphaned process.

Laffery:
## Note on Change Detection in Tests and Demos

Many test components previously did not explicitly declare a change detection strategy. Since Angular 22 defaults to `OnPush`, these components now behave differently in tests.

To avoid introducing too many changes in this PR, we have taken the following temporary measures:

- **Test components** defined in spec files are explicitly set to `changeDetection: ChangeDetectionStrategy.Eager`.
- **Demo components** used directly in tests are patched with `ɵcmp.onPush = false` in `beforeEach` to restore the previous behavior without modifying the demo source files.

These will be properly addressed in follow-up PRs.

Additionally, it should be noted that demo components also lack explicit change detection strategy declarations, meaning **all demos are now running in `OnPush` mode by default**. There is a possibility that some demos may not work correctly under this mode. This will also be addressed in subsequent PRs.

HyperLife1119 (approved):
LGTM

PR #9808, opened 2026-05-29
refactor(module:cascader,select,tree-select): clean up styles

HyperLife1119:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/NG-ZORRO/ng-zorro-antd/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Application (the showcase website) / infrastructure changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #9874, opened 2026-07-17
fix(module:select): restore virtual scroll spacer width

### LINKED ISSUE #9863: [nz-select] Wrong scroll position for the selected item when opening the dropdown
### Reproduction link
[https://stackblitz.com/edit/stackblitz-starters-aafvxapp?file=src%2Fmain.ts](https://stackblitz.com/edit/stackblitz-starters-aafvxapp?file=src%2Fmain.ts)

### Steps to reproduce
1. Create an \<nz-select\> with \`[nzDropdownMatchSelectWidth]="false"\`
2. Add more than 16 \<nz-option\> inside the select
3. Select one of the items towards the end of the list (example the 20th option)
4. Close and open the select again

### What is expected?
The dropdown is scrolled to the selected item position (20th option in our example)

The dropdown is scrolled only to the 16th option.

| Environment | Info |
|---|---|
| ng-zorro-antd | 21.3.2 |
| Browser | Chrome, Edge |

---
- Note 1: the scroll position is correct when selecting something before the 16th option (starts breaking with the 17th option)
- Note 2: removing [nzDropdownMatchSelectWidth]="false" also resolves the issue (but we would lose that nice feature then..)

Suspicion: With having >16 options in the select, virtual scroll kicks in and in combination with `[nzDropdownMatchSelectWidth]=false`, it breaks the scrolling to the selected option

https://github.com/user-attachments/assets/ce03b017-a9bd-4ec9-b117-36e956681bb8

Laffery:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

This is a regression introduced in `21.3.1` by #9808. The refactor removed the `width: 1px` declaration from the virtual-scroll spacer while retaining the `full-width` layout overrides used by `nzDropdownMatchSelectWidth=false`:

```less
.cdk-virtual-scroll-content-wrapper { position: static; }
.cdk-virtual-scroll-spacer { position: absolute; top: 0; }
```

CDK writes the spacer's full data-set height. With no width, that empty absolutely positioned spacer no longer contributes to the viewport's scrollable overflow. The viewport therefore has only the initial 16 rendered options (`8` visible options plus an `8`-item buffer) to scroll through. Reopening a Select whose selected option is beyond that range cannot position it correctly.

Issue Number: #9863

## What is the new behavior?

Restore the 1px spacer width so the CDK virtual viewport includes the complete option list in its scrollable overflow and can position selected options beyond the initial render buffer.

- [ ] Yes
- [x] No

## Other information

Validated manually with the Select large-data demo and with `npx stylelint components/select/style/patch.less` and `npm run build`.

### DIFF
--- a/components/select/style/patch.less
+++ b/components/select/style/patch.less
@@ -19,6 +19,8 @@
     .cdk-virtual-scroll-spacer {
       position: absolute;
       top: 0;
+      // Keep the spacer in the viewport's scrollable overflow area.
+      width: 1px;
     }
   }
 }


HyperLife1119 (approved):
LGTM

PR #9881, opened 2026-07-24
fix(module:tooltip,popover,popconfirm): skip hover on touch devices

roccobalocco:
…ver trigger.

Add check for touch devices before handling hover trigger.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `nzTooltipTrigger` hover behavior on devices that use touch as primary input leads to anomalies.
eg: having a button with `nz-tooltip` with `hover` trigger leads to the activation of both button action and tooltip visualization.

Issue Number: N/A

## What is the new behavior?
The new behavior avoids the activation of tooltip when a user press a component with a `nz-tooltip` on it (that use `hover` as trigger), leaving `click` trigger as it is, that is because, in my humble opinion, a touch is intended to be a click, not a hover.

- [x] Yes
- [ ] No

If this behavior was intended this PR can lead to breaking changes, impacting the behavior of a lot of web apps.
However, as a developer, I have to disable the tooltip that exploit `hover` trigger on touch devices because of this annoying behavior.

## Other information
Example to reproduce:
Imagine having something like:

```ts
<button
  nz-col
  nz-tooltip
  [nzTooltipMouseEnterDelay]="0.75"
  [nzTooltipTitle]="'search' | translate"
  nz-button
  (click)="openSearchModal = true">
  <nz-icon nzType="search" nzTheme="outline" />
</button>
```

Just press the button for enough time to satisfy the delay time and you will have (the search button is under the searchbar that has been opened):
image

To be clear, the `nzTooltipMouseEnterDelay` is what, in the example listed above, leads to the bug, but that is due to the action that I make once the searchModal is open.
There are cases in which, without any enterDelay, the tooltip fires after click and remain open as showed in the screenshot.

### DIFF
--- a/components/popconfirm/popconfirm.spec.ts
+++ b/components/popconfirm/popconfirm.spec.ts
@@ -12,10 +12,11 @@ import { delay, of, Observable } from 'rxjs';
 import { vi } from 'vitest';
 
 import { provideNzNoAnimation } from 'ng-zorro-antd/core/animation';
-import { dispatchMouseEvent } from 'ng-zorro-antd/core/testing';
+import { dispatchMouseEvent, dispatchTouchEvent } from 'ng-zorro-antd/core/testing';
 import { provideNzIconsTesting } from 'ng-zorro-antd/icon/testing';
 import { NzAutoFocusType } from 'ng-zorro-antd/popconfirm/popconfirm';
 import { NzPopconfirmModule } from 'ng-zorro-antd/popconfirm/popconfirm.module';
+import { NzTooltipTrigger } from 'ng-zorro-antd/tooltip';
 
 import { NzPopConfirmButtonProps } from './popconfirm-option';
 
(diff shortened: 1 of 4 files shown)


Laffery:
Hi @roccobalocco,

It is great to see more people contributing to NG-ZORRO—thank you for your first contribution! There are a few things that need to be addressed before we can move this forward.

1. Please update the commit message to follow our commitlint convention. This change affects the shared tooltip base and therefore impacts Tooltip, Popover, and Popconfirm. Please use a clear, valid `module:*` scope that reflects that impact (for example, `fix(module:tooltip,popover,popconfirm): ...`).

2. Please add the necessary unit tests for this behavior and make sure that all CI checks pass. The tests should cover the touch-device path as well as the existing mouse-hover behavior.

3. Most importantly, the current implementation looks a little too broad and may not cover the interaction model comprehensively. The React `rc-trigger` package and Angular Material are useful references for how mature component libraries handle touch devices, hover interactions, and compatibility mouse events. Please take a look at their approaches before refining the implementation.

Thank you again for the contribution, and feel free to start a discussion here if you would like to talk through the design.

roccobalocco:
Hi @Laffery,

Thanks a lot for the feedback! I've addressed the three points you left:

1. Updated the commit message to the correct scope (module:tooltip,popover,popconfirm), since this change touches the shared base directive.
2. Added tests in tooltip, popover, and popconfirm covering both the touch-suppression path and the existing mouse-hover behavior. In popconfirm.spec.ts I added a [nzPopconfirmTrigger] binding to the test harness to exercise the hover trigger, since it defaults to click.
3. I looked at Angular Material's tooltip ([tooltip.ts](https://github.com/angular/components/blob/main/src/material/tooltip/tooltip.ts), specifically _setupPointerEnterEventsIfNeeded()/_isTouchPlatform()), it seems to pick one of two exclusive code paths once, when listeners are first bound, and never re-evaluates. That's static enough that it doesn't adapt to hybrid devices (e.g. a touchscreen laptop with a mouse attached still permanently loses hover).

Given that, I went with per-interaction detection instead: hover keeps mouseenter/mouseleave bound unconditionally, and a short-lived flag set on touchend suppresses just the synthetic mouseenter that follows a tap, so real mouse hover on the same device still works. 

Let me know if you'd like anything adjusted or if anything was wrong.

roccobalocco:
Hey! 
Got any news?

roccobalocco:
Hi @Laffery, 
sorry to bother you again, but I was wondering if you got any news.

If I can help please let me know.

PR #9925, opened 2026-08-26
fix(module:input-number): prevent keyboard step when readonly

### LINKED ISSUE #9924: [nz-input-number]: nzReadOnly does not prevent value changes via arrow keys
### Reproduction link
[https://stackblitz.com/edit/xfqqtqzb?file=src%2Fapp%2Fapp.ts](https://stackblitz.com/edit/xfqqtqzb?file=src%2Fapp%2Fapp.ts)

### Steps to reproduce
1. Render an `<nz-input-number>` with `nzReadOnly` set to `true` and an initial value (e.g. 123).
2. Focus the control and press the ArrowUp or ArrowDown key.

### What is expected?
When the control is read-only, the value should not be changeable by any interaction. Typing is correctly blocked (the inner <input> gets the readonly attribute), so keyboard stepping should be inert as well — consistent with how a read-only form control is expected to behave.

`nzReadOnly` applies the `readonly` attribute to the inner text <input> (it blocks typing), adds the readonly CSS classes and hides the controls (up and down buttons) correctly. The step logic never consults it:

- onKeyDown() gates ArrowUp/ArrowDown stepping on `nzKeyboard()` only.
- step() early-returns only on `upDisabled()` / `downDisabled()` (min/max bounds).

Neither checks `nzReadOnly()`, so ArrowUp/ArrowDown still change the value while the control is read-only.

| Environment | Info |
|---|---|
| ng-zorro-antd | 22.0.1 |
| Browser | Chrome 151 |

sajjadBayat32:
## PR Checklist

- [x] The commit message follows our guidelines

## PR Type
- [x] Bugfix

When `nz-input-number` has `nzReadOnly` set, typing is blocked and the up/down control buttons are hidden, but pressing the ArrowUp / ArrowDown keys still changes the value. A read-only control should not be mutable through any input.

Issue Number: Fixes #9924

## What is the new behavior?
Stepping is now blocked while `nzReadOnly` is set. The guard is placed in the shared `step()` method, so it covers keyboard, handler buttons, and mouse wheel consistently.

- [x] No

## Other information
Added a unit test asserting ArrowUp / ArrowDown do not change the value when `nzReadOnly` is set.

### DIFF
--- a/components/input-number/input-number.component.spec.ts
+++ b/components/input-number/input-number.component.spec.ts
@@ -436,6 +436,18 @@ describe('input-number', () => {
     expect(component.value).toBe(0);
   });
 
+  it('should not step by keyboard when readonly', () => {
+    component.value = 5;
+    component.readonly.set(true);
+    fixture.detectChanges();
+
+    upStepByKeyboard();
+    expect(component.value).toBe(5);
+
+    downStepByKeyboard();
+    expect(component.value).toBe(5);
+  });
+
   it('should be set mouse wheel', () => {
     const input = hostElement.querySelector('input')!;
     component.value = 0;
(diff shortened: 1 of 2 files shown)


HyperLife1119 (approved):
LGTM

PR #9934, opened 2026-09-08
fix(module:icon): preserve custom aria-label

### LINKED ISSUE #9932: nz-icon: aria-label is overwritten by nzType
### Reproduction link
[https://stackblitz.com/edit/wgyfn7qo-ycavvvzt](https://stackblitz.com/edit/wgyfn7qo-ycavvvzt)

### Steps to reproduce
1. Change aria-label at nz-icon
2. Inspect the element and notice, that the aria-label is still equal the nzType value

### What is expected?
The custom aria-label should be used

The nzType is used as aria-label

| Environment | Info |
|---|---|
| ng-zorro-antd | 22.0.1 |
| Browser | every browser |

mnkprs:
Preserve a consumer-supplied `aria-label` on `nz-icon` when the icon renders or its type changes.

Fixes #9932.

## PR Checklist

- [x] The commit message follows the contribution guidelines.
- [x] Tests for the changes have been added.
- [x] Documentation is not applicable: this restores standard attribute compatibility.

## PR Type

- [x] Bugfix

`<nz-icon nzType="question-circle" aria-label="Help" />` renders with `aria-label="question-circle"`. The directive's host binding overwrites the supplied label.

## What is the new behavior?

An aliased signal input gives static and bound labels precedence over `nzType`. Empty strings are preserved; omitted, `null`, and `undefined` labels fall back to the current icon type.

Regression coverage includes both directive selectors, label and type changes, empty labels, and fallback behavior. The test fixture uses the default change detection strategy.

- [x] No

## Other information

- Rebased onto `master` at `50d1d9756`, including the CI test fix in #9939.
- Full library suite with coverage (`npm test`): 2,823 passed, 8 skipped, including all 18 icon tests in Chromium.
- Schematics suite (`npm run test:schematics`): 67 passed.
- Full lint (`npm run lint`) and TypeScript checks pass.


components/icon/doc/index.en-US.md
```diff
@@ -24,6 +24,16 @@ We keep in syncing with [antd](https://ant.design/components/icon/).
 | `[nzIconfont]`     | Type of the icon from iconfont      | `string`                       | -           | -             |
 | `[nzRotate]`       | Rotate degrees                      | `number`                       | -           | -             |
 
+### Accessibility
```
Laffery: Please remove this new Accessibility section and its Chinese counterpart. `aria-label` / `[aria-label]` are standard Angular/native attribute capabilities; this change restores compatibility with them rather than introducing an Icon-specific public API, so we should not expose them separately in the component docs.

components/icon/icon.spec.ts
```diff
@@ -249,6 +318,21 @@ export class NzTestIconExtensionsComponent {
   readonly rotate = signal(0);
 }
 
+@Component({
+  imports: [NzIconDirective],
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
Laffery: Please remove the explicit `ChangeDetectionStrategy.OnPush` here (and its now-unused import); it is already the default for this test setup.

Laffery:
Please rebase this branch onto the latest `master` and make sure the full test suite passes. The current component-test failure was not introduced by this PR and has already been fixed by an earlier PR, so rebasing should bring in the fix; no unrelated test workaround is needed here.

Laffery (approved):
The requested changes have been addressed, and the updated test suite passes. Thank you!

PR #9952, opened 2026-09-14
feat(module:core-time): support date adapter config factories

### LINKED ISSUE #9951: feat(core/time): allow factory configuration in provideNzDateFnsAdapter
## What problem does this feature solve?
Starting with v22, applications configure the date-fns adapter through `provideNzDateFnsAdapter({ locale, firstDayOfWeek })`. This works for a static locale, but it does not cover Angular applications that share one `app.config.ts` across multiple localized builds.

For example, an application built with `@angular/localize` receives its locale through the injected `LOCALE_ID`. `provideNzI18n` already supports a factory for this use case:

```ts
provideNzI18n(() => (inject(LOCALE_ID) === 'zh' ? zh_CN : en_US))
```

The equivalent is not possible with `provideNzDateFnsAdapter`, because it only accepts a configuration object. Calling `inject(LOCALE_ID)` while constructing that object happens outside an injection context. As a result, DatePicker can use the correct NG-ZORRO language pack while month and weekday names still use the date-fns default locale.

The current workaround is to provide `NZ_DATE_LOCALE` and, when needed, `NZ_DATE_CONFIG` separately with factories. This exposes lower-level tokens, splits one adapter configuration across multiple providers, and makes the helper inconsistent with `provideNzI18n`.

Supporting a factory would let the date adapter follow the localized Angular bundle automatically, keeping DatePicker month names, weekday names, and week rules aligned with `LOCALE_ID`. This is especially useful for applications that produce multiple locale-specific bundles from the same application configuration.

## What does the proposed API look like?
Allow `provideNzDateFnsAdapter` to accept either the existing configuration object or a factory that runs in an Angular injection context:

```ts
import { inject, LOCALE_ID } from '@angular/core';
import { enUS, zhCN } from 'date-fns/locale';
import { provideNzDateFnsAdapter } from 'ng-zorro-antd/core/time';

provideNzDateFnsAdapter(() => {
  const localeId = inject(LOCALE_ID);

  return {
    locale: localeId.startsWith('zh') ? zhCN : enUS,
    firstDayOfWeek: localeId.startsWith('zh') ? 1 : 0
  };
})
```

A possible type shape is:

```ts
type NzDateFnsAdapterConfigFactory = () => NzDateFnsAdapterConfig;

function provideNzDateFnsAdapter(
  config?: NzDateFnsAdapterConfig | NzDateFnsAdapterConfigFactory
): EnvironmentProviders;
```

If the implementation is shared by `provideNzDateAdapter`, the same factory support could also be exposed consistently by `provideNzNativeDateAdapter` and custom adapter providers. The existing object overload should remain unchanged.

Laffery:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The date adapter provider helpers only accept a static configuration object. Applications that derive the adapter locale from an injectable dependency such as `LOCALE_ID` have to configure the lower-level date tokens separately.

Issue Number: #9951

## What is the new behavior?
`provideNzDateAdapter`, `provideNzDateFnsAdapter`, and `provideNzNativeDateAdapter` now accept either a configuration object or a configuration factory. The factory is resolved once in an Angular injection context, so it can use `inject()` while providing both the adapter locale and date configuration.

The existing object-based API remains unchanged. Tests cover factory injection for the date-fns and native adapters, and the date adapter guide includes a localized-build example.

- [ ] Yes
- [x] No

## Other information

Validation:

- 93 targeted date adapter tests passed
- TypeScript compilation passed
- ESLint passed (existing warnings only)
- Library build passed

Closes #9951


components/core/time/date-adapter.ts
```diff
@@ -42,8 +45,26 @@ export interface NzDateAdapterConfig<TLocale = unknown> extends NzDateConfig {
  */
 export function provideNzDateAdapter<TDate, TLocale, TAdapter extends NzDateAdapter<TDate, TLocale>>(
   adapterClass: Type<TAdapter>,
-  config?: NzDateAdapterConfig<TLocale>
+  config?: NzDateAdapterConfig<TLocale> | NzDateAdapterConfigFactory<TLocale>
 ): EnvironmentProviders {
+  if (typeof config === 'function') {
```
Nicoss54: suggestion: config function is called twice, why not calling it once just after the 

```if(typeof config === 'function')```

piece of code

Comment (Laffery) [reply components/core/time/date-adapter.ts]: the place is not in injection context

Comment (Laffery) [reply components/core/time/date-adapter.ts]: I can add a memoized closure here

```ts
let resolvedConfig: NzDateAdapterConfig | undefined;
const getConfig = (): NzDateAdapterConfig => (resolvedConfig ??= config());

// then invoke `getConfig` in useFactory
```

WDYT?

HyperLife1119:
```ts
{
  provide: NZ_DATE_ADAPTER_CONFIG, // a new injection token
  useFactory: typeof config === 'function' ? config : () => config
},
{
  provide: NZ_DATE_CONFIG,
  useFactory: (): NzDateConfig => {
    const { locale: _locale, ...dateConfig } = inject(NZ_DATE_ADAPTER_CONFIG);
    return { ...NZ_DATE_CONFIG_DEFAULT, ...dateConfig };
  }
},
{
  provide: NZ_DATE_LOCALE,
  useFactory: (): TLocale | undefined => inject(NZ_DATE_ADAPTER_CONFIG).locale
}
```

This ensures that config is called only once, at the cost of requiring an injection token.

Laffery: that's great, but I don't wanna expose `NZ_DATE_ADAPTER_CONFIG` as a public injection token, I think `provideXxx` is enough and much more convenient, WDYT

HyperLife1119: Yes, NZ_DATE_ADAPTER_CONFIG is private.

components/core/time/date-adapter.ts
```diff
@@ -42,8 +45,26 @@ export interface NzDateAdapterConfig<TLocale = unknown> extends NzDateConfig {
...
+          const { locale: _locale, ...dateConfig } = config();
+          return { ...NZ_DATE_CONFIG_DEFAULT, ...dateConfig };
+        }
+      },
+      {
+        provide: NZ_DATE_LOCALE,
+        useFactory: (): TLocale | undefined => config().locale
+      }
+    ]);
+  }
```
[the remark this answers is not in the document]
Laffery: Addressed in 9cae3b8: the configuration factory is now provided through a private `NZ_DATE_ADAPTER_CONFIG` token, so Angular resolves it once in an injection context and both public tokens derive from the same result. The test now asserts a single invocation.

components/core/time/date-adapter.ts
Laffery: Updated in 9cae3b8 following the discussion: calling `config()` directly after the type guard would be outside an injection context, so the factory is now registered through a private `NZ_DATE_ADAPTER_CONFIG` token. Angular resolves it once, and `NZ_DATE_CONFIG` / `NZ_DATE_LOCALE` inject the same result. I also restored the single-invocation assertion.

Laffery: Further simplified in 1d3efa0 following the discussion: both object and factory inputs now go through the private `NZ_DATE_ADAPTER_CONFIG` provider, and the common providers derive date config and locale from it. The no-config case explicitly falls back to `{}` to keep the existing default behavior.

HyperLife1119 (approved):
LGTM

Nicoss54 (approved):
LGTM

PR #9963, opened 2026-09-18
fix(module:select): provide option context to label template

daniell0gda:
## PR Checklist

- [x] The commit message follows our guidelines

## PR Type

- [x] Bugfix

Issue Number: N/A

Reproduction: https://stackblitz.com/edit/mffaz4pk?file=src%2Fapp%2Fapp.ts

When an `nzOptions` entry uses a `TemplateRef` as its `label`, `nz-select` renders that
template in the dropdown without any context. The template therefore cannot read the
option it is being rendered for, so every item renders identically (or blank), and the
only workaround is to project `nz-option` elements and wire up `ngTemplateOutletContext`
by hand. The StackBlitz above shows both dropdowns side by side: the `nzOptions` one
falls back to the "missing context" branch, the projected `nz-option` one renders each
product.

## What is the new behavior?

The mapped option is forwarded to `nz-option-item` and exposed to the template as
`$implicit` (plus its own properties as named context values), matching how
`nzCustomTemplate` is already rendered in the top control. Templates used through
`nzOptions` can now read the option: `<ng-template #tpl let-option>{{ option.value }}</ng-template>`.

Projected `nz-option` usage is unaffected: no context is built when an option carries
none, so that path renders exactly as before.

- [ ] Yes
- [x] No

## Other information

`NzSelectItemInterface` gains an optional internal `templateContext` field. The public
`nzOptions` shape is unchanged, so no docs update was needed. Validated locally with the
select unit suite, plus `tsc --noEmit` and ESLint via the pre-commit hook.


pburgmer:
Would be nice to get this fix to solve our performance problems.

components/select/select.component.ts
```diff
@@ -704,6 +704,7 @@ export class NzSelectComponent implements ControlValueAccessor, OnInit, AfterCon
       const listOfTransformedItem = listOfOptions.map(item => {
         return {
           template: item.label instanceof TemplateRef ? item.label : null,
+          templateContext: item.label instanceof TemplateRef ? item : null,
```
Nicoss54: There is an inconsistencies in the API this way.

The new templateContext for nzOptions templates exposes raw fields (value, label, disabled), while nzCustomTemplate exposes nz-prefixed fields (nzValue, nzLabel, nzDisabled). The PR description claims it "matches how nzCustomTemplate is already rendered," but a template written against one context shape will silently break if reused for the other (e.g., option.nzValue is undefined in the nzOptions path).

components/select/option-item.component.ts
```diff
@@ -79,8 +80,15 @@ export class NzOptionItemComponent implements OnChanges, OnInit {
   @Output() readonly itemClick = new EventEmitter<NzSafeAny>();
   @Output() readonly itemHover = new EventEmitter<NzSafeAny>();
 
+  protected templateOutletContext: NzSafeAny | null = null;
+
   ngOnChanges(changes: SimpleChanges): void {
-    const { value, activatedValue, listOfSelectedValue } = changes;
+    const { value, activatedValue, listOfSelectedValue, templateContext } = changes;
+    if (templateContext) {
```
Nicoss54, line 89: Duplicate logic with the select-item.component.ts.

context-building code duplicates the existing templateOutletContext getter in select-item.component.ts:63-68 rather than sharing a helper. Minor maintenance risk since the convention now lives in two places.

Nicoss54 (changes requested):
Nice Job, needs some few changes are nitpick :)

components/select/select.component.ts
daniell0gda: please re-check

components/select/option-item.component.ts
daniell0gda: please re-check

Nicoss54 (approved):
LGTM

components/select/select.component.ts
```diff
@@ -703,6 +703,7 @@ export class NzSelectComponent implements ControlValueAccessor, OnInit, AfterCon
       const listOfOptions = this.nzOptions || [];
       const listOfTransformedItem = listOfOptions.map(item => {
         return {
+          ...item,
```
Laffery: add an optional field `templateContext` in `NzSelectItemInterface` and pass to it

components/select/select.spec.ts
```diff
@@ -963,6 +963,42 @@ describe('select', () => {
...
+      component.nzOpen.set(true);
+      await flushChanges();
+      expect(document.querySelector('nz-option-item')!.textContent?.trim()).toBe('option: test_01 star');
+    });
+
+    it('should expose the same nz-prefixed fields as nzCustomTemplate to the label template', async () => {
+      component.listOfOption.set([{ value: 'test_01', label: component.nzPrefixedOptionTemplate, disabled: true }]);
+      component.nzOpen.set(true);
+      await flushChanges();
+      expect(document.querySelector('nz-option-item')!.textContent?.trim()).toBe('option: test_01  true');
```
Laffery: why the text content is not `option: test_01 true`?


```suggestion
      expect(document.querySelector('nz-option-item')!.textContent?.trim()).toBe('option: test_01 true');
```

Laffery (changes requested):

Laffery:
nice job, however I think it should be a feature not a bugfix, it could be merged in the next minor version :)

components/select/select.spec.ts
daniell0gda: It's because when `label` is a `TemplateRef`, mapped `nzLabel` is `null`, so `{{ option.nzValue }} {{ option.nzLabel }} {{ option.nzDisabled }}` still keeps both spaces around the empty `nzLabel` → `option: test_01  true`.

Fixed.

components/select/select.component.ts
daniell0gda: Please revalidate if this is what you meant or maybe its to much as a templateContext.

